PR #4, opened 2015-03-19
Fix for clearing fact table action.

labkey-nicka:
Didn't support POSTing correctly resulting in any user being told this request has an invalid security context.


labkey-nicka:
@cnathe This is a fix for an action that isn't used all that often to clear the fact table. Main problem was that it wasn't posting back the CSRF with the form.

cnathe:
Code review complete. The change look good.

PR #7, opened 2015-04-07
Fb fixes

labkey-martyp:
Issue 22658: Apply selection when navigating.  Updated automatic promotion from selection to filter when navigating away from selection creation page.  Updated tests.

Also fixed CDSTest.verifyGrid. Failing on TeamCity.


cnathe:
Code review complete.

PR #13, opened 2015-07-16
Spec 23019: Start gutter plot for non-overlapping points

cnathe:
Instead of rendering data points that don't overlap between the x and y axis (i.e. null value in one direction) on the axis origin, render a gutter plot to the left or below the main plot for better visual relationship between the data.


cnathe:
Since the CDS tests are in a transition state for the fb_schema branch, I am going to merge this pull request to develop without waiting for TeamCity.

PR #79, opened 2016-08-08
Merge Rfb cds bug fix 16.3.1into release16.2

XingY:
This branch was created off a history point of release16.2, in order to push customer a hot fix for the June's build while release16.2 had already merged in undesirable features for July sprint. Staging server had been upgraded using a build from this branch, and client seems to be fine with the upgrade. Before merging the current 16.2 into this branch, a back up branch off this branch has been created in case clients needs more hot fixes. There has been a lot of conflicts that required manual resolving. A quick pass of manual testing has also been performed after syncing this branch with latest of 16.2.


labkey-nicka:
Code review feedback: @XingY Solid change overall. My main concern was about the use of [JavaScript Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) which is an ES6 feature. If we run into compatibility issues we can either write Set manually or a library like ImmutableJS for feature support.

PR #96, opened 2016-12-02
Merge Rfb cds bug fix into release16.3

XingY:


webapp/Connector/src/app/model/StudyProducts.js
```diff
@@ -15,6 +15,20 @@ Ext.define('Connector.app.model.StudyProducts', {
...
 
+    /**
+     * Issue 28093
+     * These "marker" products trigger special behavior because they have semantic meaning.
+     * They are not considered actual products. The expected behavior is:
+     * 1. Study learn about summary should not show NONE.
+     * 2. Study learn about detail should not show NONE.
+     * 3. Products learn about summary should not show NONE, NOT_USED, COMING_SOON
+     */
+    markerProducts: {
```
XingY: Let's define this in statics block, instead of prototype..

webapp/Connector/src/app/store/Study.js
```diff
@@ -159,14 +159,17 @@ Ext.define('Connector.app.store.Study', {
...
                     if (study.study_name === this.productData[p].study_name) {
-                        // Consider: These should probably be of type Connector.app.model.StudyProducts
-                        // but it'd be good to then have a common sourcing mechanism for LA models
-                        products.push({
-                            product_id: this.productData[p].product_id,
-                            product_name: this.productData[p].product_name
-                        });
-                        productNames.push(this.productData[p].product_name);
-                        productClasses.push(this.productData[p].product_class || '[Unknown]');
+                        if (this.productData[p].product_name.toLowerCase() !=
```
XingY: maybe also add .trim()

webapp/Connector/src/app/store/StudyProducts.js
```diff
@@ -94,6 +103,22 @@ Ext.define('Connector.app.store.StudyProducts', {
         this._onLoadComplete();
     },
 
+    getById : function(id) {
```
XingY: Can we simply push all products to product store, but use a flag showSummary to decide if the record should be shown. It would be much cleaner to have everything in store.

webapp/Connector/src/app/store/StudyProducts.js
```diff
@@ -94,6 +103,22 @@ Ext.define('Connector.app.store.StudyProducts', {
...
+        if (this.secretData[id]) {
+            return Ext.create(this.model, this.secretData[id])
+        }
+        return this.callParent(arguments);
+        // return this.secretData[id] || this.callParent(id);
+    },
+
+    validProduct : function(productName) {
```
XingY: Use Ext.iterate would be cleaner if new values are added in the future

webapp/Connector/src/app/store/StudyProducts.js
```diff
@@ -144,7 +169,13 @@ Ext.define('Connector.app.store.StudyProducts', {
                     });
                 }
                 product.other_products = otherProducts;
-                products.push(product);
+
+                if (this.validProduct(product.product_name)) {
+                    products.push(product);
+                }
+                else{
+                    this.secretData[product.product_id] = product;
```
XingY: We need to exclude "None" products from being accessible on product detail page. Also, like mentioned before, I prefer to push every product into products, but have an extra flag to indicate if it should be shown.

webapp/Connector/src/app/store/StudyProducts.js
Kristjansson: I tried the route of adding a flag, but I had a difficult time finding where to tell it not to display based on that flag.
Also, this route guarantees that these non-products are not included in filter facet views, and by anything that counts products. The way I was thinking about these (NONE, COMING_SOON, NOT_USED) is that they are semantically not products, they just happen to have the same fields. So keeping them separate from the rest of the product data just seemed prudent.

XingY:
TC has 2 random failures

PR #106, opened 2017-02-14
Rfb add study fields 29072

Kristjansson:


cnathe:
I am seeing 5 test failures on TeamCity for this feature branch and only 2 on the develop branch for release16.3. Can you verify that those failures are in fact not related to the feature branch changes?

PR #108, opened 2017-02-22
Merge Rfb cds bug fix 17.1.4 into release16.3

XingY:


XingY:
3 test failures on TC, but are either random failure or are consistent with failures on release16.3/

PR #110, opened 2017-03-03
Merge to release16.3

XingY:
Issue 29570: DataSpace: filter on categorical time point plot with alignment fails


cnathe:
We don't get TeamCity coverage on this rfb because it is release16.3.

PR #113, opened 2017-03-20
Merge Rfb group update 29387 into release17.1

XingY:


webapp/Connector/src/controller/Group.js
```diff
@@ -18,7 +18,16 @@ Ext.define('Connector.controller.Group', {
     init : function() {
 
         this.control('grouplistview', {
-            itemclick: function(v, grp) {
+            itemclick: function(v, grp, item) {
+                console.log('bar');
```
XingY: Remove console.logs

webapp/Connector/src/panel/GroupList.js
```diff
@@ -107,6 +107,10 @@ Ext.define('Connector.view.GroupListView', {
         this.callParent([config]);
 
         this.addEvents('deletegroup');
+
+        this.on('itemclick', function(grouplist, record, item){
```
XingY: This event binding doesn't seem to do anythign other than console log

webapp/Connector/src/view/GroupSummary.js
```diff
@@ -210,8 +210,22 @@ Ext.define('Connector.view.GroupSummaryBody', {
...
+                rightColumn.add({
+                    xtype: 'button',
+                    text: 'View in Plot',
+                    itemId: 'groupplotview',
+                    style: 'margin-top: 20px'
+
+                })
             }
         }
+        else
```
XingY: I would move var plotFilter to outside of if condition, so that a group will either show 'View in Plot' or 'No plot saved...'. Though unlikely to happen in real life, the current if/else grouping can end up showing neither if 'containsPlot' is true but plotFilter is null.

test/src/org/labkey/test/tests/cds/CDSGroupTest.java
```diff
@@ -183,6 +189,40 @@ public void verifyGroups()
...
+        xaxis.openSelectorWindow();
+        xaxis.pickSource(CDSHelper.TIME_POINTS);
+        xaxis.pickVariable(CDSHelper.TIME_POINTS_DAYS);
+        xaxis.confirmSelection();
+        sleep(CDSHelper.CDS_WAIT_ANIMATION);
+        yaxis.pickSource(CDSHelper.NAB);
+        yaxis.pickVariable(CDSHelper.NAB_TITERIC50);
+        yaxis.confirmSelection();
+
+        cds.saveGroup(GROUP_PLOT_TEST , "");
```
XingY: saveGroup method was intended to be used for groups created from Find Subject page only. The modification to applyAndMaybeWaitForBars resulted in a dozen test failures so I've reverted it.

test/src/org/labkey/test/tests/cds/CDSGroupTest.java
```diff
@@ -183,6 +189,40 @@ public void verifyGroups()
...
+
+        cds.clearFilters();
```
XingY: clearFilters is intended for Find Subject

PR #115, opened 2017-03-24
Fb alignment 29386

cnathe:


cnathe:
Note: this branch was incorrectly named as a "fb" instead of "rfb". After writing the new test cases, I will create a "rfb" branch off of this one to get release17.1 TeamCity coverage.

webapp/Connector/src/model/StudyAxisData.js
```diff
@@ -234,9 +240,9 @@ Ext.define('Connector.model.StudyAxisData', {
 
                 visit = study.visits[alignedDay];
 
-                if (visitTagCaption)
+                if (visitTagLabel)
                 {
-                    visit.visitTags.push(this.getVisitTag(study.name, groupLabel, visitTagCaption, groupDesc));
+                    visit.visitTags.push(this.getVisitTag(study.name, groupLabel, visitTagLabel, groupDesc));
```
XingY: Really minor comment, probably update getVisitTag function's parameter to tagLabel instead of tagCaption to avoid confusion later on.

resources/queries/cds/ds_visittag.sql
```diff
@@ -18,13 +18,9 @@ DISTINCT arm_visit.study_arm_visit_type AS Name,
...
-
-SELECT
-DISTINCT arm_visit.study_arm_visit_label AS Name,
-arm_visit.study_arm_visit_label AS caption,
-TRUE AS SingleUse
-FROM cds.import_studypartgrouparmvisit AS arm_visit
-WHERE study_arm_visit_label IS NOT NULL
\ No newline at end of file
+-- add the pre-defined alignment visits tags
+UNION ALL SELECT 'Enrollment' AS Name, 'Enrollment' AS caption, TRUE AS SingleUse
```
XingY: Is the purpose of 'UNION ALL' to fail the ETL should any of 'Enrollment', 'First Vaccination' or 'Last Vaccination' is accidentally used in StudyPartGroupArmVisit.study_arn_visit_type?

XingY: NVM, with tags from study_arm_visit_type also has False as SingleUse, this sql will never result in duplicate primary key error. Union All and Union will function just the same in this case.

cnathe:
This branch is ready to go, but the client has requested that we wait until after the March sprint build to merge this in to give them time to get the data ready on their end.

PR #119, opened 2017-05-02
Rfb learn updates 29837

Kristjansson:


webapp/Connector/src/view/Learn.js
```diff
@@ -695,7 +702,8 @@ Ext.define('Connector.view.LearnHeader', {
     getDataView : function() {
         if (!this.headerDataView) {
             this.headerDataView = Ext.create('Connector.view.LearnHeaderDataView', {
-                cls: 'dim-selector',
+                flex: 6,
```
XingY: Flex is specified for the dimension selector on the left side, but not on search field on the right, which will have no effect.

webapp/Connector/src/view/Learn.js
```diff
@@ -726,6 +734,7 @@ Ext.define('Connector.view.LearnHeader', {
             this.searchField = Ext.create('Ext.form.field.Text', {
                 emptyText: 'Search',
                 cls: 'learn-search-input',
+                width: 300,
```
XingY: Related to previous comment about missing flex config, would using flex and minWidth instead provide better user experience or is a fixed width of 300px the desired behavior?

webapp/Connector/src/app/view/StudyProducts.js
```diff
@@ -30,8 +30,9 @@ Ext.define('Connector.app.view.StudyProducts', {
         tpl: new Ext.XTemplate(
                 '<div class="detail-description">',
                     '<h2>{product_name:htmlEncode}</h2>',
-                    '<div class="detail-description-text">{product_description:htmlEncode}</div>',
-                '</div>'
+                    '<div class="detail-description-text" style="height: 42px"><p>{product_description:htmlEncode}</p></div>',
```
XingY: Can the inline style be moved to class?

PR #123, opened 2017-05-18
Merge into release17.1

XingY:


webapp/Connector/src/store/Explorer.js
```diff
@@ -7,5 +7,200 @@ Ext.define('Connector.store.Explorer', {
 
     extend: 'LABKEY.app.store.OlapExplorer',
 
-    subjectName: 'Subject'
+    subjectName: 'Subject',
+
+    //Override: Mostly the same as OlapExplorer, see code inside //DataSpace override for actual overrides
+    loadDimension : function(selections, altRequestDimNamedFilters) {
```
labkey-nicka: This is a lot of copied code for an additional line. Could consider firing an event in the original or adding a stub method to allow preprocessing of a dimension query results. I realize this would mean a change to release17.1, however, could do in hosting17.1 and run CDS off of that.

webapp/Connector/src/controller/Query.js
```diff
@@ -895,6 +895,136 @@ Ext.define('Connector.controller.Query', {
 
     encodeURLFilter : function(filter) {
         return encodeURIComponent(filter.getURLParameterName()) + '=' + encodeURIComponent(filter.getURLParameterValue());
+    },
+
+    getUserAccessibleMembersCache: function()
```
labkey-nicka: A comment here about what this members cache is for would be worthwhile.

webapp/Connector/src/store/Explorer.js
XingY: Yeah it was mostly to avoid having to triage. I prefer stub methods over event firing as the check of each each target needs to be synchronous. As for switching to hosting17.1, it's probably more overhead than to get triage on release, though it's safer..

PR #126, opened 2017-06-07
Spec Issue: 30212

Kristjansson:
- Created an initial StudyGroups.txt
 - Updated import ETL
 - Updated PopulateStudiesTask to populate permissions
 - Accounted for no-permissions and inherited permissions


src/org/labkey/cds/data/steps/PopulateStudiesTask.java
```diff
@@ -135,6 +151,99 @@ private void cleanContainers(Set<String> importStudies, Container project, User
...
+            if (!c.equals(project)) {
+                allContainerNames.add(c.getName());
+            }
+        }
+
+        for (Map.Entry<Pair<String, String>, Set<String>> entry : studyRoleGroups.entrySet())
+        {
+            Container c = ContainerManager.getChild(project, entry.getKey().getKey());
+            allContainerNames.remove(c.getName());
+            MutableSecurityPolicy policy = (new MutableSecurityPolicy(c));
```
XingY: This line creates a blank policy for each entry, which is problematic as it erases role assignment from previous entries for the same study.

src/org/labkey/cds/data/steps/PopulateStudiesTask.java
```diff
@@ -135,6 +151,99 @@ private void cleanContainers(Set<String> importStudies, Container project, User
...
+        TableInfo importStudyGroups = schema.getTable("import_studygroups");
+
+        if (null == importStudyGroups)
+            throw new PipelineJobException("Unable to find cds.import_studygroups table.");
+
+        SQLFragment sql = new SQLFragment("SELECT * FROM ").append(importStudyGroups);
+        Map<String, Object>[] importStudies = new SqlSelector(importStudyGroups.getSchema(), sql).getMapArray();
+
+        // Map((Study, Role) -> Set(Groups))
+        Map<Pair<String, String>, Set<String>> studyRoleGroups = new HashMap<>();
```
XingY: Using Map<String,Map<Role, List<UserPrincipal>>> studyRoleGroups might be easier as policy is per study.

PR #129, opened 2017-06-22
Rfb antigens variables search 30424

Kristjansson:


webapp/Connector/src/app/store/VariableList.js
```diff
@@ -16,27 +16,21 @@ Ext.define('Connector.app.store.VariableList', {
         this.callParent([config]);
     },
 
-    loadSlice : function() {
+    loadVariables : function(assayName) {
```
XingY: Not a big deal, but loadVariables is called and store is loaded every time the variable page is visited. Ideally the loadRawData should only happen once per assay.

PR #135, opened 2017-09-29
Fb r report 28752

XingY:
Merge into develop

### DIFF
--- a/test/src/org/labkey/test/tests/cds/CDSRReportsTest.java
+++ b/test/src/org/labkey/test/tests/cds/CDSRReportsTest.java
@@ -158,7 +158,8 @@ public class CDSRReportsTest extends CDSReadOnlyTest
         }
 
         goToProjectHome();
-        goToManageViews();
+        clickTab("Clinical and Assay Data");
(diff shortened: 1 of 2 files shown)


test/src/org/labkey/test/tests/cds/CDSRReportsTest.java
```diff
@@ -158,7 +158,8 @@ public void validateViewReportComplex() throws MalformedURLException
         }
 
         goToProjectHome();
-        goToManageViews();
+        clickTab("Clinical and Assay Data");
+        waitAndClick(Locator.tagWithClass("span", "labkey-wp-icon-button-active").append(Locator.tagWithClass("span", "fa-pencil")));
```
labkey-susanh: Is there a locator class that this can be added to?

XingY: It's an odd situation for both CDS and LabKey. LabKey usually uses Manage Views, that one has a helper method. But edit from manage views doesn't provide user the ability to hide report. Click edit icon by web part name does provide user option to hide and that's what I need. So it's a one-off use case. CDS rarely tests outside the application so it's weird to add a LabKey location to CDS test page also.

PR #140, opened 2017-11-17
add study_prot to all datasets

XingY:


XingY:
have 3 timeout error, shouldn't be of concern.

PR #147, opened 2018-05-09
Spec ID: 33991 - Build mAb grid UI

XingY:


test/src/org/labkey/test/tests/cds/CDSMAbGridTest.java
```diff
@@ -0,0 +1,190 @@
...
+        Locator.XPathLocator subjectInfoPane = CDSHelper.Locators.subjectInfoPaneHeader().notHidden();
+        assertElementNotPresent(subjectInfoPane);
+        CDSHelper.NavigationLink.GRID.makeNavigationSelection(this);
+        assertElementPresent(subjectInfoPane);
+        CDSHelper.NavigationLink.MABGRID.makeNavigationSelection(this);
+
+        log("Verify mAb summary grid content");
+        grid.clearAllFilters();
+        Assert.assertEquals("Number of mab/mabmix rows is not as expected", 54, grid.getMabCounts());
+        Assert.assertEquals("Geometric mean value for '2F5' is not as expected", "1.50583", grid.getMabCellValue("2F5", GEOMETRIC_MEAN_IC50_COL));
```
labkey-danield: Purely a style/subjective comment...
If I could go back and change one thing about test code I have written it would be to not assert (and possibly fail) on issues that shouldn't block checks that follow in the test.
If the number of rows is not as expected I agree we should fail, but I think it would also be useful to see if the MabCellValue of 2F5 is as expected as well. What I try to do in tests I am currently working on is setting a boolean value if an expected condition is not met, continue with the test, and then check the boolean at the end. 
I'm not saying you should change this code just something to keep in mind fo future tests.

webapp/Connector/src/view/MabGrid.js
```diff
@@ -0,0 +1,617 @@
...
+    requestExportCSV: function() {
+        this.requestExport(false);
+    },
+
+    requestExportExcel: function() {
+        this.requestExport(true);
+    },
+
+    requestExport : function(isExcel) {
+        //
```
cnathe: Is this just a placeholder for a future story?
Is the button in this case just disabled for now? If not, probably want to either disable it or give the user a message to the effect of "not yet implemented / coming soon".

webapp/Connector/src/view/MabGrid.js
```diff
@@ -0,0 +1,617 @@
...
+    },
+
+    onExport : function(isExcel) {
+        //
```
cnathe: Same as above comment.

webapp/Connector/src/window/MabGridFacet.js
```diff
@@ -0,0 +1,234 @@
...
+    initComponent : function()
+    {
+        this.callParent();
+        this.addEvents('mabfiltersearchchanged');
+        this.createFacetGrid();
+        this.on('mabfiltersearchchanged', this.facetGrid.filterFacetOptions, this.facetGrid);
+    },
+
+    // Override to add search box
+    getTopConfig : function() {
```
cnathe: If I'm reading this correctly, this search functionality only exists for the Mab facet filter window. Search seems like a valuable thing to add more generally to facet filter windows, but that is likely a future thing.

webapp/Connector/src/view/MabGrid.js
XingY: Yeah it's just a placeholder. The button is hidden for now and will be enabled in a future story.

PR #152, opened 2018-06-05
Spec ID: 34246 - Integration with/link to R reports for mAbs

XingY:


test/src/org/labkey/test/pages/cds/MAbDataGrid.java
```diff
@@ -201,6 +204,83 @@ public String getMabCellValue(String mabName, String columnLabel)
...
+    public void openIC50Report()
+    {
+        openMAbReport(NAB_MAB_IC50_REPORT);
+    }
+
+    public void openMAbReport(String reportName)
+    {
+        _webDriverWrapper.click(Locators.getMAbReportBtn(reportName));
+        _webDriverWrapper.sleep(500);
+        _webDriverWrapper.waitForElement(Locators.getMAbReportHeader(reportName));
```
labkey-danield: If you would rather you could try removing the sleep on the line before and add:
WebDriverWrapper.WAIT_FOR_JAVASCRIPT +  500 
as the second parameter to the waitForElement.

PR #169, opened 2018-09-21
Spec ID: 35220 - mAb Learn About Pages

XingY:


resources/schemas/dbscripts/postgresql/cds-18.22-18.23.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE cds.import_MAbMixMetadata ADD COLUMN mab_mix_type VARCHAR(250);
+ALTER TABLE cds.import_MAbMixMetadata ADD COLUMN mab_mix_name_other VARCHAR(250);
+ALTER TABLE cds.import_MAbMixMetadata ADD COLUMN mab_mix_lanlid VARCHAR(250);
+ALTER TABLE cds.MAbMixMetadata ADD COLUMN mab_mix_type VARCHAR(250);
+ALTER TABLE cds.MAbMixMetadata ADD COLUMN mab_mix_name_other VARCHAR(250);
+ALTER TABLE cds.MAbMixMetadata ADD COLUMN mab_mix_lanlid VARCHAR(250);
+ALTER TABLE cds.mAbMetaGridBase ADD COLUMN mab_mix_type VARCHAR(250);
+
+CREATE TABLE cds.studymAb (
```
labkey-nicka: Should `prot` and `mab_mix_id` be `NOT NULL` as well?

webapp/Connector/src/app/model/MAb.js
```diff
@@ -0,0 +1,56 @@
...
+        {name: 'mab_mix_name_std'},
+        {name: 'mab_mix_type'},
+        {name: 'mab_mix_lanlid'},
+        {name: 'mab_mix_lanlid_link'},
+        {name: 'mab_mix_name_other'},
+        {name: 'other_labels'},
+
+        {name: 'has_data'},
+
+        {name: 'mabs', convert : function(value) {
```
labkey-nicka: Could declare a converter function that all of the model fields could use throughout the app that are attempting to ensure Array. There may be other convert methods that could be shared as well?

```js
function asArray(value) {
    return Ext.isArray(value) ? value : [];
}

// usage
{name: 'some_field', convert: asArray}
```

webapp/Connector/src/app/store/MAb.js
```diff
@@ -0,0 +1,179 @@
...
+                mixRec.donors = donors;
+                mixRec.donors_str = donors.join(', ');
+                mixRec.isotypes = isotypes;
+                mixRec.isotypes_str = isotypes.join(', ');
+                mixRec.hxb2Locs = hxb2Locs;
+                mixRec.hxb2Locs_str = hxb2Locs.join(', ');
+                mixRec.mabnames_str = mabnames.join(', ');
+
+                mabs = Object.values(mabIdMap);
+                mabs.sort(function(a, b) {
```
labkey-nicka: We should use `Ext.Object.getKeys` here in lieu of `Object.values` since [IE does not support Object.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values#Browser_compatibility).

```js
mabs = Ext.Object.getKeys(mapIdMap).sort(/* ... */);
```

Should probably scan the app for other usages as well.

webapp/Connector/src/app/store/MAb.js
```diff
@@ -0,0 +1,179 @@
...
+                    return a.mabName - b.mabName;
```
labkey-nicka: Natural Sort?

```js
mabs.sort(function(a, b) {
    return LABKEY.internal.SortUtil.naturalSort(a.mabName, b.mabName);
});
```

PR #181, opened 2018-12-06
Spec ID: 35945 - mAb Saved Groups

XingY:


webapp/Connector/src/view/GroupSave.js
```diff
@@ -730,9 +734,35 @@ Ext.define('Connector.view.GroupSaveList', {
...
+        var isMab = this.isMab;
+        this.store.filterBy(function(record) {
+            if (isMab)
+                return record.get('type') === 'mab';
+            else
+                return record.get('type') !== 'mab';
+        });
+    },
+
+    cloneGroupStore (source) {
```
labkey-bradh: This errors out on IE, and the app breaks. Browser complains of missing colon

PR #194, opened 2019-03-18
Epic 4474: PK Data in Plot/Main App 

XingY:


test/src/org/labkey/test/pages/cds/XAxisVariableSelector.java
```diff
@@ -126,9 +130,88 @@ public void setCellType(String... value)
...
+    {
+        _test.mouseOver(Locator.tagWithText("div", optionLabel + ":"));
+        String xpathDimField = getTimeOptionAdvancedOptionXPathDimField(optionLabel);
+        Locator.XPathLocator locDimField = Locator.xpath(xpathDimField);
+        if (_test.isElementPresent(locDimField))
+        {
+            WebElement element = locDimField.findElement(_test.getDriver());
+            if (element != null)
+                return element.getText();
+        }
```
labkey-tchad: Use `locDimField.findElementOrNull` to find elements that may not be present to avoid finding an element multiple times; In this case, when the element is present, the test will find the element once during `isElementPresent` then again when you call `findElement` to actually get the WebElement.

test/src/org/labkey/test/pages/cds/DataspaceVariableSelector.java
```diff
@@ -149,6 +152,19 @@ public void pickVariable(String variable)
         _test.sleep(CDSHelper.CDS_WAIT_ANIMATION);
     }
 
+    public List<String> getAvailableVariables()
+    {
+        List<String> variables = new ArrayList<>();
+        Locator.XPathLocator selectorPanelLoc = Locator.tagWithClass("div", "variable-selector");
+        Locator.XPathLocator variableLoc = selectorPanelLoc.append(Locator.tagWithClass("div", "content-label").notHidden());
+        List<WebElement> variableElements = variableLoc.findElements(_test.getDriver());
+        if (variableElements != null)
```
labkey-tchad: This null check is unnecessary, `findElements` can't return `null`

test/src/org/labkey/test/pages/cds/XAxisVariableSelector.java
```diff
@@ -126,9 +130,88 @@ public void setCellType(String... value)
...
+        List<String> options = new ArrayList<>();
+        String xpathDimField = getTimeOptionAdvancedOptionXPathDimField(optionLabel);
+        String xpathDimDropDown = "//div[contains(@class, 'advanced-dropdown')][not(contains(@style, 'display: none'))]";
+
+        Locator.XPathLocator locDimField = Locator.xpath(xpathDimField);
+
+        if (_test.isElementPresent(locDimField))
+        {
+            _test.longWait().until(LabKeyExpectedConditions.animationIsDone(locDimField));
+            _test.click(locDimField);
```
labkey-tchad: As above, use `findElementOrNull` for elements that might not be present. In this case, the test will find the element three times.

PR #204, opened 2019-06-13
Fixes for CDS on trunk

labkey-klum:


webapp/Connector/src/utility/MabQuery.js
```diff
@@ -540,8 +540,7 @@ Ext.define('Connector.utility.MabQuery', {
     },
 
     _getProcessedIC50Ranges : function(f) {
-        var value = Ext.isArray(f.getValue()) ? f.getValue()[0] : f.getValue();
-        var ranges = value.split(';');
+        let ranges = f.getValue();
```
cnathe: Are we using let / const in the cds JS code? If so, should this be const?

webapp/Connector/src/utility/Query.js
```diff
@@ -862,8 +862,8 @@ Ext.define('Connector.utility.Query', {
                         var operator = f.getFilterType().getURLSuffix();
                         if (operator == 'eq' || operator == 'in')
                         {
-                            var strvalue = Ext.isArray(f.getValue()) ? f.getValue()[0] : f.getValue();
-                            listofFilters.push(operator=='eq' ? [strvalue] : strvalue.split(';'));
+                            let value = f.getValue();
```
cnathe: Are we using let / const in the cds JS code? If so, should this be const?

webapp/Connector/src/window/MabGridFacet.js
```diff
@@ -106,9 +106,9 @@ Ext.define('Connector.window.MabGridFacet', {
         var filter = this.mabModel.getFieldStateFilter(fieldName);
 
         if (filter) {
-            var value = Ext.isArray(filter.getValue()) ? filter.getValue()[0] : filter.getValue();
+            let value = filter.getValue();
```
cnathe: Are we using let / const in the cds JS code? If so, should this be const?

PR #209, opened 2019-07-09
Fb public announcement page

labkey-klum:


README.md
```diff
@@ -0,0 +1,53 @@
+# CDS
```
cnathe: I think you should get Xing to review this README once she is back. I don't have much context on this anymore.

src/org/labkey/cds/view/template/FrontPage.jsp
```diff
@@ -62,10 +62,43 @@
...
-        }
+        };
+
+        LABKEY.Utils.onReady(function ()
+        {
+            $(document).ready(function(){
+
+                // notification close button
+                $('div.dismiss').click(function(){
+                    $('#notification').remove();
```
cnathe: This removes it from the UI, but do we need to call the core-dismissCoreWarnings API at all from the FrontPage.jsp like we do for Header.js below?

webapp/Connector/src/view/Header.js
```diff
@@ -119,3 +119,86 @@ Ext.define("Connector.view.Header", {
...
+                if (o.messages) {
+                    this.setVisible(true);
+                    this.msg.update(o.messages);
+                }
+                else
+                    this.setVisible(false);
+            }
+        }, this);
+
+        this.callParent();
```
cnathe: I would usually put this `this.callParent()` line above the Ajax request (so move it up to line 170), but if this works, you could keep it here.

webapp/Connector/src/view/Header.js
```diff
@@ -119,3 +119,86 @@ Ext.define("Connector.view.Header", {
...
+            method: 'GET',
+            scope : this,
+            success: function (response){
+                var o = Ext.decode(response.responseText);
+                if (o.messages) {
+                    this.setVisible(true);
+                    this.msg.update(o.messages);
+                }
+                else
+                    this.setVisible(false);
```
cnathe: I assume this is not needed since the component has a default state of `hidden: true`.

webapp/Connector/src/view/Header.js
```diff
@@ -119,3 +119,86 @@ Ext.define("Connector.view.Header", {
...
+            '<div class="notification-messages">',
+            '<span>{.}</span><br>',
+            '</div>',
+            '</tpl>'
+        );
+
+        return new Ext.Component({
+            tpl : tpl,
+            flex : 2,
+            margin : '15, 5, 15, 5'
```
cnathe: I don't think I have seen this syntax for the margin prop. I would have expected it to accept a single int value for the case where all sides are the same or a string like the following: `margin: '15px 5px'` for this case.

src/org/labkey/cds/view/template/FrontPage.jsp
labkey-klum: probably not, since there is no session established at this point

PR #214, opened 2019-08-20
Issue 38213: Update header text in MabGrid.js.

labkey-bpatel:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38213


labkey-bpatel:
test failures are unrelated to the header text update, merging..

PR #215, opened 2019-08-22
Issue 37456: MAb mix label as color - no line color

labkey-bpatel:
-Remove isDemographic check so that the lines in line plot get colored for non-demographic fields.
-Remove fields from Color selector as per related ticket 37338.


labkey-bpatel:
@labkey-sweta - Can you please verify the fixes/updates in this fb. Let me know if you have questions.

PR #219, opened 2019-09-16
Epic 6087: Inclusion of Non-Integrated Datasets

labkey-bpatel:


XingY:
is this targeting develop or release19.2snapshot?

labkey-bpatel:
targeting develop

webapp/Connector/src/app/view/module/DataAvailabilityModule.js
```diff
@@ -142,7 +146,21 @@ Ext.define('Connector.view.module.DataAvailabilityModule', {
                     {
                         ftype: 'grouping',
                         collapsible: false,
-                        groupHeaderTpl: new Ext.XTemplate('{name:htmlEncode}')
+                        groupHeaderTpl: new Ext.XTemplate('{name:htmlEncode}',
+                                '<tpl if="name === \'MAb Administration Studies\'">',
```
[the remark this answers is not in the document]
labkey-bpatel: @XingY - I added it here since MAb learn page has 'Integrated Data' section (previously called 'Data Availability'), and the grouping based on study_type happens here.  Any idea as to where this instruction could be added if not here?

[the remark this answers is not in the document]
XingY: I like how this.data.instructionsForAdministration is passed in instead of hardcoded. Would the following modification to cube.js allow pure data driven tpl? 

hasGrouping: true,
                            groupingSubHeaders: {
                                'MAb Administration Studies' : 'Go to Plot to view or Grid to export.  Additional non-integrated data files may be available for download. See study page.',
                                'MAb Characterization Studies' : 'Go to Plot to view or Grid to export.  Additional non-integrated data files may be available for download. See study page.'
                            },

webapp/Connector/src/app/store/Study.js
```diff
@@ -94,6 +99,11 @@ Ext.define('Connector.app.store.Study', {
         this._onLoadComplete();
     },
 
+    onLoadLearnAssays : function(learnAssayData) {
+        this.learnAssayData = learnAssayData.rows;
```
XingY: not sure why learnAssayData is needed since assayData already contains assay information. It's a left join from studyassay to assay, so should already be able to tell if a assay page exist.

webapp/Connector/src/app/store/Study.js
```diff
@@ -341,27 +356,64 @@ Ext.define('Connector.app.store.Study', {
...
                 }).length > 0;
 
-                //non-integrated assay that's in cds.assays, document, and studydocument
+                //non-integrated assay with potentially downloadable data, which may or may not also have a learn assay page
                 study.non_integrated_assay_data = documents.filter(function (doc) {
                     return doc.label && doc.docType === 'Non-Integrated Assay';
                 });
 
-                //non-integrated assay that's in studyassay
+                //non-integrated assay that has subject metadata in cds.studyassay, which may or may not also have a learn assay page
```
XingY: remove "subject" from comment, cds.studyassay doesnt contain subject metadata, just metadata

webapp/Connector/src/app/store/Study.js
```diff
@@ -341,27 +356,64 @@ Ext.define('Connector.app.store.Study', {
...
-                        isLinkValid: undefined,
-                        suffix: undefined,
-                        sortIndex: undefined,
-                        filePath: undefined,
-                        hasPermission: niAssay.has_access,
-                        assayIdentifier: niAssay.data_id,
-                        hasAssayData: niAssay.has_data
-                    };
-                    study.non_integrated_assay_data.push(nonIntegratedAssay);
+
```
XingY: a more efficient way would be to iterate nonIntegratedAssays once to build a map<assayIdentifier, nonIntegratedAssay>

webapp/Connector/src/app/store/Study.js
```diff
@@ -134,6 +158,9 @@ Ext.define('Connector.app.store.Study', {
                 return study.study_name;
             });
 
+            this.assayIdentifiers.integratedAssays = this.assayIdentifiers.map(integratedAssay => integratedAssay.assay_identifier);
```
XingY: use a new variable instead of assigning result to this.assayIdentifiers.integratedAssays

webapp/Connector/src/app/store/Study.js
```diff
@@ -330,6 +366,71 @@ Ext.define('Connector.app.store.Study', {
...
+
+                    //combine when there is assay page and downloadable data and subject metadata
+                    var niToCombine = study.non_integrated_assay_data.filter(function(documentData) {
+                        return documentData.assayIdentifier === niAssay.data_link_id || documentData.assayIdentifier === niAssay.data_id;
+                    });
+                    var indexToRemove = study.non_integrated_assay_data.findIndex(function(documentData) {
+                        return documentData.assayIdentifier === niAssay.data_link_id || documentData.assayIdentifier === niAssay.data_id;
+                    });
+                    if (indexToRemove >= 0) {
+                        study.non_integrated_assay_data.splice(indexToRemove, 1);
```
XingY: instead of assign, splice then re push, why not just use a separate variable to hold non_integrated_assay_data.

webapp/Connector/src/app/store/Study.js
labkey-bpatel: what about the data that is in Assay but not in StudyAssay (our initial implementation), in which case assayData won't have the info we need.

webapp/Connector/src/app/store/Study.js
labkey-bpatel: good suggestion, much cleaner!
done!

webapp/Connector/src/app/store/Study.js
labkey-bpatel: done

webapp/Connector/src/app/store/Study.js
labkey-bpatel: this is now refactored with a cleaner approach.

webapp/Connector/src/app/store/Study.js
labkey-bpatel: done!

webapp/Connector/src/app/store/Study.js
XingY: I don't think that's a legit scenario. A study page will only list assays defined in studyassay relationship. StudyDocument is to provide detailed info for the relationship, but it doesn't define the relationship.

labkey-bpatel:
Ok to merge, only one unrelated test failure on teamcity - https://teamcity.labkey.org/viewLog.html?buildId=831595&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95

PR #222, opened 2019-09-20
Migrate module deps

ilikepies101:


build.gradle
```diff
@@ -4,6 +4,7 @@ dependencies
 {
     BuildUtils.addLabKeyDependency(project: project, config: "compile", depProjectPath: ":server:optionalModules:dataintegration", depProjectConfig: "apiCompile")
     BuildUtils.addLabKeyDependency(project: project, config: "jspCompile", depProjectPath: ":server:optionalModules:dataintegration", depProjectConfig: "apiCompile")
+    BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "study"), depProjectConfig: "published", depExtenstion: "module")
```
labkey-susanh: Module.properties declares a dependency on dataintegration as well, which seems relevant since it's also a compile dependency.

PR #226, opened 2019-10-10
Develop fb tours

jmtaylor-fhcrc:


labkey-klum:
Hi Jason, I'm going to create a new feature branch in our repo an pull over your changes so that we can get TeamCity test coverage on the FB. Most of the styling changes look fine and what I can do is update the way that the tours are both applied to the page and the UI elements on the home page so that it compatible with the approach of sourcing the content from a wiki. We can discuss in more detail at the meeting next week.

PR #227, opened 2019-10-11
Epic # 6212 Home page updates

labkey-bpatel:


src/org/labkey/cds/CDSModule.java
```diff
@@ -70,6 +85,10 @@
     final ModuleProperty _mabReport2ID;
     final ModuleProperty _mabReport2Label;
 
+    //wiki properties for the front page
+    final ModuleProperty _whatYouNeedToKnowWiki;
+    final ModuleProperty _takeATourWiki;
+
```
labkey-klum: @labkey-bpatel 
Do these need to be member variables? They don't appear to be used elsewhere in this class.

webapp/Connector/src/component/Started.js
```diff
@@ -22,52 +22,71 @@ Ext.define('Connector.component.Started', {
...
+                                    '<tr><td>{[this.getTakeATourWiki()]}</td></tr>',
+                                '</table></div>',
+                            '</td>',
+                            '<td class="tile-detail" style="padding-left:10%"><h3 class="tile-title">Try it out</h3>',
+                                '<div><table>',
+                                    '<tr><td class="tile-detail-static"><a href=\'#learn/learn/Study/\' target="_blank">Learn about</a> studies, products, assays, antibodies, ad publications</td></td></tr>',
+                                    '<tr><td class="tile-detail-static"><a href=\'#summary\' target="_blank">Find subjects</a> with common characteristics</td></td></tr>',
+                                    '<tr><td class="tile-detail-static"><a href=\'#chart\' target="_blank">Plot</a> assay results across studies and years of research</td></td></tr>',
+                                    '<tr><td class="tile-detail-static">Compare <a href=\'#mabgrid\' target="_blank">Monoclonal antibodies</a> and their neutralization curves</td></td></tr>',
+                                '</table></div>',
```
labkey-klum: @labkey-bpatel 
Did the spec say to open these links in another tab (I didn't notice it)? Seems like we would just want to navigate within the app.

src/org/labkey/cds/CDSModule.java
labkey-bpatel: They don't need to be member variables, however, there are other ModuleProperty variables above it declared similarly so thought to follow the same pattern.

webapp/Connector/src/component/Started.js
labkey-bpatel: Ah ok! wasn't too clear on the default behavior so went with opening in another tab. 
But now updated - clicking on those links will open the page in the same tab.

labkey-bpatel:
@labkey-klum - tours related changes looks good!

labkey-bpatel:
Teamcity test verification:
Teamcity is mostly happy, except for one test failure in CDSMAbTest.verifyMabAndSubjectGroups related to 'Curated groups and plots' -- which is also happening on develop and unrelated to this feature work.

fb: https://teamcity.labkey.org/viewLog.html?buildId=838876&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95

develop: https://teamcity.labkey.org/viewLog.html?buildId=838326&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95

PR #231, opened 2019-10-17
Allow filtering of variables in the dimension selector portion of the axis selectors

labkey-klum:


labkey-klum:
I think I misinterpreted what Drienna wanted in this feature. We don't need to distinguish between x or y axis selectors but we do need to distinguish between the measures and dimensions in the axis selectors.

PR #236, opened 2019-11-21
Epic 5788 - Show all for long lists 

labkey-bpatel:


webapp/Connector/src/app/store/MAb.js
```diff
@@ -144,9 +144,31 @@ Ext.define('Connector.app.store.MAb', {
...
+                }
+
                 Ext.each(mstudies, function(studymab) {
                     var study_id = studymab.prot;
+
+                    //get index of a group/study_type (so far there are just the two groups: 'MAb Characterization Studies' & 'MAb Administration Studies')
+                    var grpIdx = groupWithIndexes.findIndex(function(value) {
+                       return value ? value.groupName === studymab.study_type : false;
+                    });
+
```
labkey-klum: isn't it possible that grpIdx will be -1, and if so will it cause problems downstream?

labkey-bpatel: Hi Karl - In this case, it doesn't look likely given the structure of the data (explanation below), but it makes sense to make the code more robust, so further check is now added. 

(Since the data (stored in mstudies variable) is a result of learn_mabStudies.sql (where study_type is the group), there will be either zero resulting rows or at least one group present at minimum. If it's zero rows then it won't enter that for loop (line 164). Also, groupWithIndexes is constructed from mstudies in the first place, which will have at least one group if there's data in mstudies.)

labkey-bpatel:
Good point about having the threshold defined in one place, where would be a good place to add such constant?

labkey-klum:
There doesn't seem to be an obvious place, maybe create a new singleton class in the constant folder. There are existing objects there : ModuleViewsLookup, State, Templates

PR #240, opened 2019-12-19
Epic 6611: PK MAB key fields need to be adjusted

labkey-bpatel:


webapp/Connector/src/model/ChartData.js
```diff
@@ -457,7 +461,12 @@ Ext.define('Connector.model.ChartData', {
                 xname: xa.label,
                 yname: ya.label,
                 colorname: ca.label,
-                rowKey: rowKey
+                rowKey: rowKey,
+                querySrc: ya.query,
```
XingY: This is too much hardcoding. 
- Since allowForGrouping is already defined at measure level, it should iterate through measures to find the fields needed for grouping, instead of hardcode here again for PKMAb. This allows other assay to define their own line group as well. 
- a single [groupField] can be built and passed to plot, instead of having plot rebuild the group value.

webapp/Connector/measure.js
```diff
@@ -635,7 +635,9 @@ Ext.define('Connector.measure.Configuration', {
                     isRecommendedVariable: true
                 },
                 'study_PKMAb_mab_mix_id': {
-                    hiddenInPlot: true
+                    hiddenInPlot: true,
+                    allowForGrouping: true,
```
XingY: The other 3 fields should also have this property set. I understand they are not strictly needed since they are included in plot data as default dimensions. But having explicit annotation makes it clear which fields are for grouping:
study_PKMAb_lab_code
study_PKMAb_source_assay
study_PKMAb_specimen_type

webapp/Connector/src/model/Measure.js
```diff
@@ -38,6 +38,7 @@ Ext.define('Connector.model.Measure', {
         {name: 'requiredInGrid', type: 'boolean', defaultValue: false},
         // used for fields hidden from advanced options but available on X, color & grid
         {name: 'hiddenInAdvancedOptions', type: 'boolean', defaultValue: false},
+        {name: 'allowForGrouping', type: 'boolean', defaultValue: false},
```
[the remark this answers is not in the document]
XingY: Also suggest to change to isLinePlotGroupingField

webapp/Connector/src/model/ChartData.js
```diff
@@ -429,10 +429,14 @@ Ext.define('Connector.model.ChartData', {
             if (_row[QueryUtils.STUDY_ALIAS] && _row[QueryUtils.VISITROWID_ALIAS]) {
                 timeAxisKey = ChartUtils.studyAxisKeyDelimiter + xVal;
                 timeAxisKey += ChartUtils.studyAxisKeyDelimiter + _row[QueryUtils.STUDY_ALIAS];
-                if (_row[QueryUtils.TREATMENTSUMMARY_ALIAS]) {
-                    timeAxisKey += ChartUtils.studyAxisKeyDelimiter + _row[QueryUtils.TREATMENTSUMMARY_ALIAS];
+                if (ya.query === "PKMAb" && _row["study_PKMAb_mab_mix_id"]) {
+                    timeAxisKey += ChartUtils.studyAxisKeyDelimiter + _row["study_PKMAb_mab_mix_id"];
```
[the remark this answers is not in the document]
labkey-bpatel: It doesn't - that was my initial (& ineffective) attempt but missed to revert it after other more effective changes were added.
Now reverted!

webapp/Connector/src/view/Chart.js
```diff
@@ -1036,7 +1036,7 @@ Ext.define('Connector.view.Chart', {
                 return row.y;
             },
             group: function(row) {
-                return row['subjectId'];
+                return row.querySrc === "PKMAb" ? (row['subjectId'] + row['mab_mix_id'] + row['source_assay'] + row['lab_code'] + row['specimen_type']) : row['subjectId'];
```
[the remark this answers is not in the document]
labkey-bpatel: Great idea, now done!

webapp/Connector/src/model/Measure.js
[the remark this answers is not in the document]
labkey-bpatel: done!

webapp/Connector/src/model/ChartData.js
labkey-bpatel: Great suggestion, now done.

webapp/Connector/measure.js
labkey-bpatel: Adding 'isLinePlotGroupingField' to these fields came in handy in constructing groupField, so this was a good addition overall.

webapp/Connector/src/model/ChartData.js
```diff
@@ -451,17 +452,26 @@ Ext.define('Connector.model.ChartData', {
...
-                    grouping += "-" + _row[m.measure.alias]
+            Ext.each(measureSetAliases, function(measureAlias) {
+                var measureRecord = Connector.getService('Query').getMeasureRecordByAlias(measureAlias);
+
+                if (measureRecord && measureRecord.data.isLinePlotGroupingField && _row[measureAlias]) {
+                    grouping += "-" + _row[measureAlias]
                 }
             }, this);
 
+            // for cases when color fields are excluded from the grouping done above
```
XingY: can this be moved to inside the Ext.each iterator above?

PR #243, opened 2020-01-17
Fb visualization test investigation

labkey-sweta:


labkey-klum (approved):
should we make the timeout for both of these tests to be less than 55 minutes? Maybe update them to be 30 minutes apiece.

PR #249, opened 2020-01-28
Item 6709: CDS bug fixes related to ETL, report lag, treatment filter, url encoding, & buttons display at different screen resolutions

labkey-bpatel:
Ticket 39344: ETL error message - cvd264 in studyrelationship table
Ticket 39119: Lag opening report when clicking on mAb report buttons
Ticket 39104: Active filters not applying treatment
Ticket 39351: mAb 'MVN/A12' link defaults to 'MVN' mAb page
Ticket 39120: MAb right most report button hidden at different zoom levels


webapp/Connector/src/view/MabGrid.js
```diff
@@ -720,6 +726,18 @@ Ext.define('Connector.view.MabGrid', {
             return false;
         }
 
+        // Fix for Issue 39229: Lag opening report when clicking on mAb report buttons
+        // Immediately go to the report view before prepareMAbReportQueries step below instead of staying on the grid view
+        // This report view is a temporary one, and will be later removed
```
XingY: Can Connector.getQueryService().prepareMAbReportQueries be moved to mabreportview itself so we don't need to do the temp view thing?

XingY (approved):
Would be good if mabreportview itself handles loading, fine to leave it as it is.

webapp/Connector/src/view/MabGrid.js
labkey-bpatel: Agreed! now done!

XingY (approved):

webapp/Connector/src/model/Measure.js
```diff
@@ -47,6 +47,9 @@ Ext.define('Connector.model.Measure', {
         {name: 'isHoursType', type: 'boolean', defaultValue: false},
         {name: 'allowHoursTimePoint', type: 'boolean', defaultValue: false},
 
+        // Property to skip displaying in plot tooltip
```
XingY: "Property" -> "Array of properties"

webapp/Connector/src/utility/PlotTooltip.js
```diff
@@ -197,8 +197,12 @@ Ext.define('Connector.utility.PlotTooltip', {
     buildPointAxisDetailTooltip: function(plot, axis, record, aggregators, hierarchicalDimensionInfo) {
         var content = '<div class="axis-details">', colon = ': ', linebreak = '<br/>';
         if (axis) {
+            var skipToolTip = axis.skipToolTip;
             var dimensions = axis.options.dimensions;
             for (var dim in dimensions) {
+                if (skipToolTip && skipToolTip.length > 0 && skipToolTip.indexOf(dim) >= 0) {
```
XingY: Could use Ext.isArray to check instead, string also has length and indexOf, which might be confusing

XingY (approved):
Just some minor comments

webapp/Connector/src/model/Measure.js
labkey-bpatel: Done!

webapp/Connector/src/utility/PlotTooltip.js
labkey-bpatel: Excellent point! Done!

PR #263, opened 2020-03-13
Epic 6919 : Update virus metadata in Learn

labkey-bpatel:
- Update AssayNAbAntigenMetadata tables & ETL script.
- Create AntigenPanelMetadata tables & ETL script.
- Create VirusPanel tables & ETL script.
- Update Antigen tab to include new columns, and minor styling changes.
- Feature Request 39428: Modify home page to clarify for HIV


resources/queries/cds/nabAntigenWithPanelMeta.sql
```diff
@@ -0,0 +1,40 @@
...
+ag.clade,
+ag.antigen_description,
+ag.antigen_control,
+ag.virus_full_name,
+ag.virus_name_other,
+ag.virus_species,
+ag.virus_host_cell,
+ag.virus_backbone,
+ag.cds_virus_id,
+GROUP_CONCAT(DISTINCT ap.panel_name, ';') AS panel_names
```
XingY: can panel_name itself contain ';'? if so, the ui can't split it correctly.

resources/etls/LoadApplication.xml
```diff
@@ -144,6 +144,14 @@
             <source schemaName="cds" queryName="ds_bamaantigen" />
             <destination schemaName="cds" queryName="BAMAAntigen" targetOption="truncate" />
         </transform>
+        <transform id="PopulateAntigenPanel" type="org.labkey.di.pipeline.TransformTask">
```
XingY, line 147: I know we are using this pattern of import table -> ds_ query -> target table a lot, but I believe the ds_antigenPanel and ds_virusPanel queries are not necessary since they are identical to their import_ counterparts

resources/schemas/dbscripts/postgresql/cds-20.001-20.002.sql
```diff
@@ -0,0 +1,87 @@
...
+TRUNCATE --necessary since cds_virus_id currently has null values and can't otherwise make it part of Primary Key below. Data will get repopulated as part of ETL run after server upgrade.
+    cds.nabantigen
+RESTART IDENTITY CASCADE;
+
+ALTER TABLE cds.nabantigen DROP CONSTRAINT nabantigen_pkey;
+ALTER TABLE cds.nabantigen ADD PRIMARY KEY (container, assay_identifier, antigen_name, target_cell, antigen_type, cds_virus_id);
+
+CREATE TABLE cds.import_virusPanel
+(
+    rowid SERIAL,
```
XingY: we haven't used rowid for cds. is there a reason to use them here? Why can't pk just be (container, cds_virus_id, cds_panel_id)? Since cds tables are repopulated each time etl is re-run, easiest is to not using identity column. I do see import_NABMAb table has a row_id, i believe the reason is part of the actual key contains a float data type which is not accepted PK so row_id is used as db pk as workaround

resources/queries/cds/nabAntigenWithPanelMeta.sql
labkey-bpatel: Hmm good point. I am using least likely one "|" instead - what do you think?
From handful of sample data client had mentioned, it looked like a pretty straight forward text.

resources/etls/LoadApplication.xml
labkey-bpatel: Now removed, and using the table names instead.

resources/schemas/dbscripts/postgresql/cds-20.001-20.002.sql
labkey-bpatel: Ah got it! Now removed!

XingY (approved):

labkey-bpatel:
Two unrelated & same test failures as the default branch. Merging!

PR #275, opened 2020-06-08
Secure Issue 40593: User not prompted for registration survey after password setting

labkey-bpatel:
#### Rationale
NeedSurvey property was determined using lastLogin being null, but it is no longer the case. Also, cds.appAction is not a mutating action so is unable to set the NeedSurvey value.

#### Changes
- Create a new MutatingApiAction that sets NeedSurvey value to true after newly registered user sets the password.

### DIFF
--- a/src/org/labkey/cds/CDSController.java
+++ b/src/org/labkey/cds/CDSController.java
@@ -227,11 +227,26 @@ public class CDSController extends SpringActionController
     private static final String ANALYTICS_USER_GROUP = "Active CAVD Member";
 
     @RequiresNoPermission
+    public class UpdateNeedSurveyAction extends MutatingApiAction<Object>
+    {
+
+        @Override
+        public Object execute(Object o, BindException errors) throws Exception
+        {
+            CDSManager.get().setNeedSurvey(getUser(), true);
+
+            ApiSimpleResponse response  = new ApiSimpleResponse();
+            response.put("success", !errors.hasErrors());
+            return response;
+        }
+    }
+
+    @RequiresNoPermission
     @IgnoresTermsOfUse
     public class AppAction extends SimpleViewAction
     {
         @Override
-        public ModelAndView getView(Object o, BindException errors) throws Exception
+        public ModelAndView getView(Object o, BindException errors)
         {
             Container c = getContainer();
             StudyService sss = StudyService.get();
@@ -250,8 +265,6 @@ public class CDSController extends SpringActionController
             }
             else if (AuthenticationManager.isRegistrationEnabled() && (needSurvey || CDSManager.get().isNeedSurvey(getUser()) || getUser().getLastLogin() == null)) // on first log in, flag user as need survey
             {
-                if (needSurvey || getUser().getLastLogin() == null)
-                    CDSManager.get().setNeedSurvey(getUser(), true);
                 String surveyParam = getViewContext().getActionURL().getParameter("survey");
                 if (surveyParam == null || !"true".equals(surveyParam))
                 {
--- a/src/org/labkey/cds/CDSController.java
+++ b/src/org/labkey/cds/CDSController.java
@@ -226,7 +226,7 @@ public class CDSController extends SpringActionController
 
     private static final String ANALYTICS_USER_GROUP = "Active CAVD Member";
 
-    @RequiresNoPermission
+    @RequiresPermission(ReadPermission.class)
     public class UpdateNeedSurveyAction extends MutatingApiAction<Object>
     {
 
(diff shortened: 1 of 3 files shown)


src/org/labkey/cds/CDSController.java
```diff
@@ -226,16 +226,33 @@ public void setUserProperties(JSONObject userProperties)
 
     private static final String ANALYTICS_USER_GROUP = "Active CAVD Member";
 
+    @RequiresNoPermission
```
XingY: should require read permission for updates

src/org/labkey/cds/CDSController.java
```diff
@@ -226,16 +226,33 @@ public void setUserProperties(JSONObject userProperties)
...
     public class AppAction extends SimpleViewAction
     {
         @Override
-        public ModelAndView getView(Object o, BindException errors) throws Exception
+        public ModelAndView getView(Object o, BindException errors)
         {
             Container c = getContainer();
             StudyService sss = StudyService.get();
             boolean hasStudy = null != sss && null != sss.getStudy(c);
+            boolean needSurvey = Boolean.parseBoolean(getViewContext().getActionURL().getParameter("NeedSurvey"));
```
XingY: I don't think this is needed with the ajax, right?

webapp/frontPage/js/modal.js
```diff
@@ -395,14 +402,25 @@ define(['jquery', 'magnific', 'util'], function($, magnific, util) {
...
+
+            //sets NeedSurvey to true
+            $.ajax({
+              url: LABKEY.ActionURL.buildURL("cds", "updateNeedSurvey.api"),
+              method: 'POST',
+              data: {
+                'X-LABKEY-CSRF': LABKEY.CSRF
+              }
+            }).success(function() {
+              window.location = LABKEY.ActionURL.buildURL("cds", "app.view", null, {NeedSurvey:true}); // set password should log user in automatically
```
XingY: instead of setting NeedSurvey to true, which would trigger another redirect in controller, could instead use param "survey=true", which is actually what prompts survey page.

XingY (changes requested):

src/org/labkey/cds/CDSController.java
labkey-bpatel: I had to add this specifically because  line 266, CDSManager.get().isNeedSurvey(getUser()) was returning false. Didn't think it was needed either, esp. since this code is executed as part of success callback after setting NeedSurvey to true.

webapp/frontPage/js/modal.js
labkey-bpatel: done!

src/org/labkey/cds/CDSController.java
labkey-bpatel: This is now fixed by OntologyManager to get NeedSurvey value as we discussed.

src/org/labkey/cds/CDSController.java
labkey-bpatel: Done!

XingY (approved):
looks great

XingY:
After buddy testing, I wonder if it's better to still let user proceed to app with "Unable to set NeedSurvey property to true" error. We probably shouldn't stop them from accessing site if they can't fill out survey due to tech issue. Admin can tell user wasn't able to complete survey from site logs. Also if they refresh the page, they will be redirected to the app anyways, since they are logged in. But if they are presented with this error, they might try to click "Create Account" button again, which is gonna say "unable to... account already created.." error, which is confusing.

But otherwise, buddy testing looks good.

labkey-bpatel:
@XingY - User can now access the site if they can't fill out survey due to tech issue.

webapp/frontPage/js/modal.js
```diff
@@ -395,14 +402,26 @@ define(['jquery', 'magnific', 'util'], function($, magnific, util) {
...
+              method: 'POST',
+              data: {
+                'X-LABKEY-CSRF': LABKEY.CSRF
+              }
+            }).success(function() {
+              window.location = LABKEY.ActionURL.buildURL("cds", "app.view", null, {survey:true}); // set password should log user in automatically
+            }).error(function(e) {
+              var errorMsg = 'Unable to set NeedSurvey property to true';
+              createAccountError(e, errorMsg);
+              window.location = LABKEY.ActionURL.buildURL("cds", "app.view");
```
XingY, line 417: maybe put redirect in a setTimeout call so user can briefly see the error (or maybe skip error at all since user won't make any sense of the msg)

labkey-bpatel: Good point! i've added error to console.log instead for dev troubleshooting.

XingY: Good call. Another (one last) minor change: use console.error instead of console.log.

XingY (approved):

webapp/frontPage/js/modal.js
labkey-bpatel: Done!

PR #281, opened 2020-06-24
Add GitHub release workflow definitions

labkey-tchad:


labkey-susanh (approved):
I doubt the test failures have anything to do with this branch

PR #294, opened 2020-09-02
Ticket 40760: treatment and assay schemas - add log messages for troubleshooting

labkey-bpatel:
#### Rationale
Treatment and Assay schema links not showing on staging server, and unable to repro the issue locally, so adding log messages for the action in question for troubleshooting.

#### Changes
- Add log messages for investigation on ValidateStudySchemaLinkAction.
- Fix display for Groups and Methods which contains html tags.

### DIFF
--- a/src/org/labkey/cds/CDSController.java
+++ b/src/org/labkey/cds/CDSController.java
@@ -1049,14 +1049,14 @@ public class CDSController extends SpringActionController
 
             boolean isValidLink = false;
             WebdavService service = ServiceRegistry.get().getService(WebdavService.class);
-            LOG.info("File path: " + Path.parse(form.getFilename()).toString());
+            LOG.info("File path: " + Path.parse(form.getFilename()).toString()); //TODO: Added for investigating Ticket 40760, to be removed.
             WebdavResource resource = service.lookup(Path.parse(form.getFilename()));
 
             if (resource != null)
             {
                 File requestedFile = resource.getFile();
-                LOG.info(requestedFile.getName() + " is present: " + (requestedFile != null));
-                LOG.info(requestedFile.getName() + " is readable: " + requestedFile.canRead());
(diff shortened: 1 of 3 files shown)


XingY (commented):
Please add TODO comment to remove those logging

src/org/labkey/cds/CDSController.java
```diff
@@ -1041,18 +1041,22 @@ else if (isPost())
...
             WebdavService service = ServiceRegistry.get().getService(WebdavService.class);
-            WebdavResource resource = service.lookup(Path.parse(form.filename));
+            LOG.info("File path: " + Path.parse(form.getFilename()).toString()); //TODO: Added for investigating Ticket 40760, to be removed.
+            WebdavResource resource = service.lookup(Path.parse(form.getFilename()));
 
             if (resource != null)
             {
                 File requestedFile = resource.getFile();
+                LOG.info(requestedFile.getName() + " is present: " + (requestedFile != null));//TODO: Added for investigating Ticket 40760, to be removed.
+                LOG.info(requestedFile.getName() + " is readable: " + requestedFile.canRead());//TODO: Added for investigating Ticket 40760, to be removed.
```
XingY: if requestedFile is null, this would throw NPE.

XingY (changes requested):
Looking at how this was implemented, it's suspicious to me that LABKEY.contextPath was prepended to href of those files on the client, but the filename (webdav path) param for the validation api does not have the LABKEY.contextPath. The prod/stage server has contextPath at root. Maybe before pushing a debug build to staging, you can try setting up your dev instance to have root context to see if it repros.

labkey-bpatel:
No repro locally even with contextpath change to "root" from 'labkey'.

src/org/labkey/cds/CDSController.java
labkey-bpatel: yes, good point, now updated.

XingY (approved):

PR #295, opened 2020-09-09
Propagate virus metadata updates throughout application

labkey-bpatel:
#### Rationale
We recently updated the AssayNAbAntigen_Metadata table and the antigens tab on the Learn – Assays NAb and NAb MAb pages to include additional virus metadata fields. We now want to propagate those additional fields in other places of the application.

#### Changes
- Update Nab & NABMAb schemas
- Update virus ETL query for NAb & MAb
- Schema updates to remove virus metadata columns
- Update NAb Grid column selector
- Include additional virus metadata in a specific order - in x-axis, color, NAb Grid, NAb & MAb Export, NAb tooltip in the plot
- Replace virus name with full virus name - in tooltip on hover over virus name in Plot variable selector, MAb grid, MAb Info pane


webapp/Connector/src/grid/Panel.js
```diff
@@ -129,19 +129,23 @@ Ext.define('Connector.grid.Panel', {
...
-                                if (ah.toLowerCase() === Connector.studyContext.subjectColumn.toLowerCase() || ah.toLowerCase() === Connector.studyContext.subjectLabel.toLowerCase())
-                                    return -1;
-                                else if (bh.toLowerCase() === Connector.studyContext.subjectColumn.toLowerCase() || bh.toLowerCase() === Connector.studyContext.subjectColumn.toLowerCase())
-                                    return 1;
-
-                                // sort columns alphabetically by title
-                                return ah == bh ? 0 : (ah > bh ? 1 : -1);
-                            })
+                    columns: value.sort(function(a, b) {
+                        if (isMeasure && a.measure.isVirusMetadataField && b.measure.isVirusMetadataField){
```
XingY: This sort function doesn't seem quite right to me, though it probably does work for the specific scenarios you need, for other coincidental reasons. Consider the following 3 objects:
A: {isVirusMetadataField: true, sortOrder:100, columnName: colA}
B: {isVirusMetadataField: true, sortOrder:1, columnName: colZ}
C: {z, isVirusMetadataField: false, columnName: colC}

 compare A and B returns 1 (A > B), compare A and C returns -1 (A < C), compare B and C returns 1 (B > C). So the sort function is not transitive.

XingY (approved):
Just a comment on the sort function not being complete. otherwise looks good.

webapp/Connector/src/grid/Panel.js
XingY: On manual testing I find that indeed this sort function doesn't work and I'm not getting ordered columns on grid as expected, probably due to its broken transitivity.

XingY (changes requested):

webapp/Connector/src/grid/Panel.js
labkey-bpatel: This should be now fixed, please review and test again. Thanks, Xing!

XingY (approved):

labkey-bpatel:
TC is almost happy with 1 unrelated test failure. Merging!

PR #299, opened 2020-09-26
Update target cell source for NAb and NAbMAb

labkey-bpatel:
#### Rationale
1. Feature Request 40348: Propagate virus metadata updates throughout application: 
- Target cell should not be coming from the metadata table, therefore, should not be removed from the AssayNAb and AssayNAbMAb tables.
- Target cell was originally in the virus metadata table, but turns out its really metadata about the assay, not the virus. Values for target cell in the metadata table at this point should be blank.

#### Changes
- Add target_cell column back to NAb and NAbMab tables & update target cell source to be not from the metadata table.
- Drop target_cell from nab antigen metadata tables.
- Remove target_cell from metadata & nabantigen query.
- Remove comments added to troubleshoot 'Support Ticket 40760: treatment and assay schemas - show link only when file exists'


resources/schemas/dbscripts/postgresql/cds-20.005-20.006.sql
```diff
@@ -0,0 +1,7 @@
+-- add target_cell columns back since data for target_cell should be coming from nab and nabmab assays, and not nabantigen metadata table
+ALTER TABLE cds.import_nab ADD COLUMN target_cell VARCHAR(250);
+ALTER TABLE cds.import_nabmab ADD COLUMN target_cell VARCHAR(250);
+
+-- Drop NOT NULL constraint since target_cell column is now expected to have null vals in nabantigen metadata table
+ALTER TABLE cds.import_nabantigen ALTER COLUMN target_cell DROP NOT NULL;
```
XingY: Since target_cell is from assay itself, should it be removed from nabantigen? Or is it used somewhere else?

webapp/Connector/src/plugin/DocumentValidation.js
```diff
@@ -73,7 +73,13 @@ Ext.define('Connector.plugin.DocumentValidation', {
         LABKEY.Ajax.request({
             url: LABKEY.ActionURL.buildURL("cds", "validateStudySchemaLink.api"),
             params: {
-                filename: schema_link
+
+                // Support Ticket 40760: Treatment and assay schemas link not showing up on Learn/Study pages
+                // Treatment and Assay schema files reside at '/_webdav/home/files/@files/static/',
+                // however, this link is stored as '/_webdav/home/files/%40files/static/' in cds.study, where '@' is already encoded to %40;
```
XingY: Sounds like Drienna should fix the study.txt file, instead of this fix

resources/schemas/dbscripts/postgresql/cds-20.005-20.006.sql
labkey-bpatel: Drienna had mentioned that "Values for target cell in the metadata table at this point should be blank." -- not too sure if there will be data in the future, hence have kept it as is for now.

webapp/Connector/src/plugin/DocumentValidation.js
labkey-bpatel: Ah ok, thought that those paths were stored in that manner (with %40 instead of @) for some valid reason, but if data can be changed then that would be great! I'll update the ticket requesting that.

labkey-bpatel: This change is now reverted, and ticket is updated to request the change in data.

resources/schemas/dbscripts/postgresql/cds-20.005-20.006.sql
labkey-bpatel: target_cell now removed, OK'd by client.

XingY (approved):

PR #300, opened 2020-10-01
Minor cds test fixes

labkey-danield:
#### Rationale
Fix some of the less than reliable test automation.

#### Related Pull Requests
* None.

#### Changes
* Added better checks/waits for UI components.
* Changed to use the correct mouseOver call for the pop-up test in Learn About.


labkey-bpatel (commented):
Thank you for these improvements, they look great!

labkey-bpatel (approved):

PR #350, opened 2021-04-24
42807: Links to study page need to prompt user for login rather than take users to the front page

labkey-klum:
#### Rationale
Currently, if a guest clicks on a link like : https://dataspace.cavd.org/cds/CAVD/app.view#learn/learn/Study/vtn505?q=HVTN%20505 they will be redirected to the public front page. Dataspace would like the experience changed so that the login dialog is rendered and if/and after the user logs in they will be redirected to the app page on the URL.

#### Related Ticket
- https://www.labkey.org/Dataspace/Feature%20Requests/issues-details.view?issueId=42807

#### Changes
Modify the modal initialization on the front page to allow configured modals to show if there is a hash value. There is no validation of the hash route and the login modal is the only one which allow this.

### DIFF
--- a/webapp/frontPage/js/application.js
+++ b/webapp/frontPage/js/application.js
@@ -73,7 +73,8 @@ require(['jquery', 'scroll', 'modal', 'util'], function( $, scroll, modal, util)
     initializeModals: function() {
       modal.initialize({
         name: 'signin-modal',
-        query_param_regex: /login=true|returnUrl=/i
+        query_param_regex: /login=true|returnUrl=/i,
+        show_for_page_nav: true       // render modal if there is a page route on the url
       });
 
       modal.initialize({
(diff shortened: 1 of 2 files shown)


labkey-klum:
@XingY you would be good to review this. There are other modals on the page which can be automatically activated through special query params. It would be good to check whether this change breaks any existing scenarios.

XingY (approved):
The change looks good for redirect to front page with login modal for CDS pages. But from the ticket, it seems to me the client might also expect links that would go to LKS (for cds, mostly it's webdav static file links) to redirect to front page with log in modal open. 
https://dataspace.cavd.org/login/login.view?returnUrl=%2F_webdav%2Fhome%2Ffiles%2F%2540files%2Fstatic%2Fvtn505_treatment_schema.html%3F

We should probably clarify in the ticket that file links will redirect to LKS log in page, if that's not their expectation.

PR #353, opened 2021-04-29
Integrated Data subheader links tooltips

labkey-martyp:
#### Rationale
Added Plot, Grid, Monoclonal Antibodies links back (as implemented by Binal) with fixed tooltips

#### Changes
* Uncomment addition of links
* Add classes to links and filter those links out when applying mouseover actions


labkey-bpatel (approved):
Looks great!

PR #364, opened 2021-05-17
Bind the enter key to the register account form.

labkey-klum:
#### Rationale
Fixes this issue:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43179

#### Changes
The registration modal previously did not bind the enter key, so if a user wanted to use keyboard navigation and submit the form after filling in the kaptcha it wouldn't have the same effect as clicking on the submit button. This change uses the same pattern as some of the other modals to bind the enter key to the form submission button.

### DIFF
--- a/webapp/frontPage/js/modal.js
+++ b/webapp/frontPage/js/modal.js
@@ -168,6 +168,16 @@ define(['jquery', 'magnific', 'util'], function($, magnific, util) {
           }
         });
       }
+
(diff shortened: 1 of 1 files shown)


labkey-bpatel (approved):
Looks good! 
Verified as well - clicking on Enter now shows the same behavior as clicking on Register button

PR #372, opened 2021-06-03
CDSVisualizationTest test fix.

labkey-sweta:
#### Rationale
Adding the missing setShowHiddenVariables and resetShowHiddenVariables
#### Related Pull Requests
* 

#### Changes
test/src/org/labkey/test/tests/cds/CDSVisualizationTest.java - 2 methods added.


labkey-danield (commented):
Does this need to go into 21.3 as well?
Change is fine otherwise, just want to make sure it is going into the correct branch.

labkey-danield (approved):

PR #377, opened 2021-06-10
Increase margin for the frontpage welcome message

labkey-klum:
#### Rationale
This issue : https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43339 reports that the welcome message gets clipped by the registration div when the window height gets too short.

image

This isn't a problem on higher resolution screens but could be problematic with legacy desktop screens or if the window was sized too short.

#### Changes
- Introduce a mixin : legacy-desktop which is triggered on window sizes shorter than 864px
- Increase the welcome div margin from 2.67 to 10.67em when the threshold is triggered
- Note that there was currently a tablet mixin, but that rule was triggered when the window width was less than 700px

### DIFF
--- a/theme/front-page/index/sections.scss
+++ b/theme/front-page/index/sections.scss
@@ -67,7 +67,10 @@
     overflow: hidden;
     margin: 2.67em 0;
     @include tablet {
-      margin: 3em 0em 2.67em 0;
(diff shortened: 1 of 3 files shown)


labkey-bpatel (approved):
looks good!
also, verified it locally

PR #378, opened 2021-06-11
43362: Outage notice out of place on CDS login page

labkey-klum:
#### Rationale
If configured (from site settings) the ribbon message divs are styled incorrectly: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43362

#### Changes
During the conversion from Sencha Cmd to build the CDS stylesheets, the CSS for these divs did not get ported over, this change re-introduces the previous styles.

### DIFF
--- a/theme/front-page/application.scss
+++ b/theme/front-page/application.scss
@@ -7,3 +7,4 @@
 @import 'index/modal';
 @import 'index/sections';
 @import 'index/overrides';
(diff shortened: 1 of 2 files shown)


labkey-bpatel (approved):
Looks good!
Verified locally, and the "outage notice" part renders very briefly, it is hardly noticeable, but its no longer sticking around.

PR #408, opened 2021-09-14
21.7 fb explore study tooltips

labkey-klum:
#### Rationale
In order to give users more information about studies in both the info pane and explorer bars, we want to display tooltips (using the hopscotch callouts) showing the study or treatment arm descriptions when hovering over the active filters panel or explorer bars for either studies or assays.

https://docs.google.com/document/d/1nLX4BeKGJB6ED-RVqxb--hyaDjwTs9VIFGx84TvXb7E/edit#

#### Changes
- source study descriptions from cds.study and treatment arm descriptions from cds.treatmentarm tables
- display the descriptions on the active filters panel for studies and treatments
- display the descriptions on the explorer bars for assays and studies, the bars need to be expanded to display the study or treatment arm
- add a utility to store descriptions on the client side


webapp/Connector/src/view/SingleAxisExplorer.js
```diff
@@ -191,6 +191,64 @@ Ext.define('Connector.view.SingleAxisExplorer', {
...
+
+                if (tooltip) {
+                    var displayTooltip = setTimeout(function() {
+                        calloutMgr.createCallout(Ext.apply({
+                            id: id,
+                            arrowOffset: 20,
+                            yOffset: 0,
+                            showCloseButton: false,
+                            target: highlighted[0],
+                            placement: 'top',
```
labkey-bpatel, line 216: Usually i've seen the placement of the tooltip on the right (or left), in this case, shouldn't it be on the right? If so, then  xOffset & yOffset will need to be updated as well

labkey-klum: For the explorer bars it was difficult to place it precisely to the right of the bar text so that it wouldn't obscure the text, it was easier and more consistent to place it over the bar. Also if we tried to place it to the right, depending on the length of the text, the tooltip position would jump around which seemed sub optimal. I think normally we would place it to the left but since that wasn't possible above was the next best option.

labkey-bpatel (approved):

PR #414, opened 2021-10-04
show loading mask for learn pages : assay, study, publications, products

labkey-klum:
#### Rationale
Some of the learn pages on the CAVD production server are slow to load due to the amount of data that must be queried for and loaded. To help improve the user experience we will add loading masks to the following pages:
- study
- assay
- products
- publications

These are the pages which are the most query intensive.

#### Changes
Update the base `LearnSummary` class to render an optional loading mask. The page must fire a `dataloaded` event to dismiss the loading mask. The mask is configured to wait for 250 ms before rendering.


labkey-bpatel (approved):
Looks good! Lot of the files were from study tooltip story as well, which ive glanced at again. Also, thanks for handling tooltip wording and styling for FR 42306

PR #422, opened 2021-12-06
21.11 fb data added filter

labkey-klum:
#### Rationale
Now that we have the publication data and non-integrated data availability implemented, we should consider updating the Data Added column filter. Currently, the user selects a number from a list (i.e. the number of assays). This has always been a little awkward and it doesn't account for publication or non-integrated data.

Associated feature request:
- https://www.labkey.org/Dataspace/Feature%20Requests/issues-details.view?issueId=43741

#### Changes
Instead of showing the data availability counts in the filter, we will display the following options for each learn page:
- Study : data types available (integrated, non-integrated, publication, none) and data available (data added, data not added)
- Assay : data types available
- Products : data types available
- MAbs : data types available
- Publications : data types available

### DIFF
--- a/webapp/Connector/src/app/store/SavedReports.js
+++ b/webapp/Connector/src/app/store/SavedReports.js
@@ -41,7 +41,27 @@ Ext.define('Connector.app.store.SavedReports', {
         this._onLoadComplete();
     },
 
-    _onLoadComplete : function() {
+    _onLoadComplete: function () {
         // override and implement
+    },
+
+    getDataTypesAvailable: function (rec) {
+        var available_data_types = [];
+        if (rec.data_availability)
+            available_data_types.push('Integrated data');
+        if (rec.ni_data_availability)
+            available_data_types.push('Non-integrated data');
+
+        if (rec.publications) {
+            var pubDataAvailable = rec.publications.filter(function (pub) {
+                return pub.available_data_count > 0;
+            })
+            if (pubDataAvailable.length > 0)
+                available_data_types.push('Publication data');
+        }
+        if (available_data_types.length === 0)
+            available_data_types.push('Data not added');
+
+        return available_data_types;
     }
 });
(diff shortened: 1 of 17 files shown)


webapp/Connector/src/app/view/Study.js
```diff
@@ -45,9 +45,13 @@ Ext.define('Connector.app.view.Study', {
         resizable: false,
         dataIndex: 'assays_added_count',
```
XingY, line 46: I've noticed an edge case with sorting on Data Added column. This is not a new issue, but with the change of filter from "# of Assay Added" to "Data Available", the issue has been made more obvious. The problem is, the sort is operated on 'assays_added_count', but the column displays other counts as well, such as 'publication count'. I think ideally, the sort should sort on "# of assay added", and if equal (or both null), then sort on other data types, such as sum of 'non-integrated added' and 'Publication data'. 

If we care to change the sorting behavior, a workaround I can think of is to add a new 'study_data_added_sort_ind' column, which is calculated as (assays_added_count*100 + non_integrated_count + publication_count).

Production site, sorted by 'Data Added":
Screen Shot 2021-12-07 at 11 56 28 AM


Local:
Screen Shot 2021-12-07 at 11 56 40 AM

webapp/Connector/src/app/store/SavedReports.js
```diff
@@ -41,7 +41,27 @@ Ext.define('Connector.app.store.SavedReports', {
         this._onLoadComplete();
     },
 
-    _onLoadComplete : function() {
+    _onLoadComplete: function () {
         // override and implement
+    },
+
+    getDataTypesAvailable: function (rec) {
```
XingY, line 48: It's fine to leave as is, but this seems can also be a static method

XingY (approved):

webapp/Connector/src/app/view/Study.js
labkey-klum: Thanks Xing, this is a good point. I'll plan to bring this up with the team during our next CDS meeting.

PR #434, opened 2022-01-08
44607: Update CDS login page to always encode error messages

labkey-klum:
#### Rationale
Always encode any error messages, and if the error contains optional adviceText and adviceHref properties, display those properly:

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=44607


labkey-adam (approved):
Looks familiar 😄

PR #450, opened 2022-02-25
Attempting to fix some intermittent failures

labkey-tchad:
#### Rationale
There are numerous race conditions in the CDS tests. This is an attempt to fix the ones that we're seeing frequently these days. It also fixes some real bugs that have been concealed by all of the test noise.

#### Changes
* Fix button locators to exclude hidden buttons
* Update virus data file names to match what the server is looking for
* **Server**: Make `TSVCopyConfig` case sensitive even when the file system is not
* Centralize and improve tooltip dismissal
* Modernize `MAbDataGrid` a bit
* Update helper to set module properties via API
* Make axis checks less strict


labkey-klum (approved):
server changes look good.

labkey-sweta (approved):
Thank you

PR #453, opened 2022-03-08
44561: Updates to virus metadata tables

labkey-klum:
#### Rationale
Iteration on schema design for the virus metadata tables

#### Changes
- drop unique constraint on `virus_metadata_all.virus_full_name`
- redefine PK for `virus_lab_id` to : `(cds_virus_id, lab_code, lab_virus_id, container)`
- redefine PK for `virus_synonym` to : `(cds_virus_id, virus_synonym, container)`

### DIFF
--- a/resources/etls/LoadApplication.xml
+++ b/resources/etls/LoadApplication.xml
@@ -251,17 +251,23 @@
             <source schemaName="cds" queryName="import_assayReport"/>
             <destination schemaName="cds" queryName="assayReport" targetOption="truncate"/>
         </transform>
+        <transform id="TruncateVirusLabId" type="org.labkey.di.pipeline.TransformTask">
+            <destination schemaName="cds" queryName="virus_lab_id" targetOption="truncate"/>
+        </transform>
+        <transform id="TruncateVirusSynonym" type="org.labkey.di.pipeline.TransformTask">
+            <destination schemaName="cds" queryName="virus_synonym" targetOption="truncate"/>
+        </transform>
         <transform id="PopulateVirusMetadataAll" type="org.labkey.di.pipeline.TransformTask">
             <source schemaName="cds" queryName="import_virus_metadata_all"/>
             <destination schemaName="cds" queryName="virus_metadata_all" targetOption="truncate"/>
         </transform>
         <transform id="PopulateVirusLabId" type="org.labkey.di.pipeline.TransformTask">
             <source schemaName="cds" queryName="import_virus_lab_id"/>
-            <destination schemaName="cds" queryName="virus_lab_id" targetOption="truncate"/>
+            <destination schemaName="cds" queryName="virus_lab_id"/>
         </transform>
         <transform id="PopulateVirusSynonym" type="org.labkey.di.pipeline.TransformTask">
             <source schemaName="cds" queryName="import_virus_synonym"/>
-            <destination schemaName="cds" queryName="virus_synonym" targetOption="truncate"/>
+            <destination schemaName="cds" queryName="virus_synonym"/>
         </transform>
         <!--Should occur last-->
         <transform id="ClearAndWarmCube" type="TaskrefTransformStep">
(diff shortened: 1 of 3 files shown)


resources/schemas/dbscripts/postgresql/cds-22.003-22.004.sql
```diff
@@ -0,0 +1,42 @@
...
+ALTER TABLE cds.virus_synonym DROP CONSTRAINT PK_VirusSynonym;
+
+-- set columns in the compound PK to not nullable
+ALTER TABLE cds.import_virus_lab_id ALTER COLUMN lab_code SET NOT NULL;
+ALTER TABLE cds.import_virus_lab_id ALTER COLUMN lab_virus_id SET NOT NULL;
+ALTER TABLE cds.virus_lab_id ALTER COLUMN lab_code SET NOT NULL;
+ALTER TABLE cds.virus_lab_id ALTER COLUMN lab_virus_id SET NOT NULL;
+
+ALTER TABLE cds.import_virus_synonym ALTER COLUMN virus_synonym SET NOT NULL;
+ALTER TABLE cds.virus_synonym ALTER COLUMN virus_synonym SET NOT NULL;
```
labkey-bpatel, line 35: Not sure whether these tables have rows with null values for these columns - if so then would need to truncate the tables (since ETLs will have to be run anyway to populate these tables), otherwise it will throw an error during server startup.

resources/etls/LoadApplication.xml
```diff
@@ -251,17 +251,23 @@
...
             <destination schemaName="cds" queryName="virus_metadata_all" targetOption="truncate"/>
         </transform>
         <transform id="PopulateVirusLabId" type="org.labkey.di.pipeline.TransformTask">
             <source schemaName="cds" queryName="import_virus_lab_id"/>
-            <destination schemaName="cds" queryName="virus_lab_id" targetOption="truncate"/>
+            <destination schemaName="cds" queryName="virus_lab_id"/>
         </transform>
         <transform id="PopulateVirusSynonym" type="org.labkey.di.pipeline.TransformTask">
             <source schemaName="cds" queryName="import_virus_synonym"/>
-            <destination schemaName="cds" queryName="virus_synonym" targetOption="truncate"/>
```
labkey-bpatel, line 264: Just curious, why remove truncate here and add a separate transform task above?

labkey-bpatel (approved):

resources/schemas/dbscripts/postgresql/cds-22.003-22.004.sql
labkey-klum: none of the sample data import tables have NULL values, and I did confirm with Jayson that these other fields are okay to make non-nullable, but that's a good point and I'll remind him to do that on staging. I'm guessing the staging tables are empty since he was running into the previous PK constraint violations during import.

resources/etls/LoadApplication.xml
labkey-klum: Both the `virus_lab_id` and `virus_synonym` tables have FK to the `virus_metadata_all` table. I found that I need an initial step to clear out those tables first before populating metadata and eventually re-populating those tables. I didn't know if there is a way to do this in a single ETL transform, but if there are standard ways to do this, I'm happy to use that pattern. I'm guessing this is a fairly common scenario.

resources/schemas/dbscripts/postgresql/cds-22.003-22.004.sql
labkey-bpatel: I was thinking to add Truncate statements in the upgrade script, but asking Jason to truncate works too.

resources/etls/LoadApplication.xml
labkey-bpatel: Got it, that makes sense, was just curious, thanks for clarifying

PR #458, opened 2022-03-17
22.3 fb issue 45053

labkey-klum:
#### Rationale
This fixes a handful of CDS tests which were failing reliably:

- `CDSGroupTest.verifySavedGroupAfterDataChange` - CDS reload ETL failed.
  - The actual ETL fix to truncate those tables was committed in 21.11 and now merged into this branch. There was also a problem with the test due to using the initial response from the command to check the status after the pipeline job. However that status is not updated after the initial launch of the job and will usually still report `QUEUED` even though the job completed successfully.
- `CDSTestLearnAbout.testNonIntegratedData` - Two tooltips displayed
  - There was a potential race condition with the tooltip that would probably only occur in an automation scenario where the hover tip would render after we had navigated and the original element it was supposed to bind to no longer existed.  
- `CDSVisualizationPlotTest.verifyPKLinePlot` - Line should also use color for a non-Subject Characteristic field that won't change over time expected:<27> but was:<7>
  - This was caused by changes to the test data, the plot was correct, but the colors applied to the categories were different.


labkey-klum:
@labkey-sweta could you review the test changes and @labkey-bpatel if you could look at the server changes. Thanks!

test/src/org/labkey/test/tests/cds/CDSVisualizationPlotTest.java
```diff
@@ -650,7 +650,7 @@ public void verifyPKLinePlot() throws Exception
         coloraxis.confirmSelection();
 
         plotGreenLineCount = getElementCount(greenLineLoc);
-        assertEquals("Line should also use color for a non-Subject Characteristic field that won't change over time", 27, plotGreenLineCount);
+        assertEquals("Line should also use color for a non-Subject Characteristic field that won't change over time", 7, plotGreenLineCount);
```
labkey-sweta, line 653: Did something change in the product? How come the expected number reduced?

labkey-sweta (approved):
Approving the test part with a question

test/src/org/labkey/test/tests/cds/CDSVisualizationPlotTest.java
labkey-klum: The test data changed with the addition of the LLOQ fields. Overall the plot is correct : 34 subjects with 27 in one category and 7 in the other. It's just that with the new fields causing a shift in the category color there are 27 black and 7 green lines instead of 27 green and 7 black lines. Took me a while to figure it out until Cory and Xing helped me.

labkey-bpatel (approved):

PR #475, opened 2022-05-27
Updates to data export button. (archived in SC)

labkey-klum:
#### Rationale
As part of this [feature request](https://docs.google.com/document/d/1QWDYnx_4BdCDxpMvLq3SWYW35DJqXAZExojGYg892aE/edit), we want to consolidate the export buttons for CSV and Excel formats into a single menu style button. We will be planning to use this button when data export support is added to the learn pages but for now this story is just to create the new component and wire it into the existing MAb and data grids.

image

#### Changes
- add a new export button component
- styling to match specifications
- replace existing MAb and data grid buttons with the new button
- fix existing automated tests

### DIFF
--- a/theme/connector/lib/ext-02.css
+++ b/theme/connector/lib/ext-02.css
@@ -6228,7 +6228,6 @@ x-boundlist-item {
     from(#fb4770),
     to(#fb2e5c)
   );
-  border-radius: 9px;
 }
 
 .x-menu-item-active .x-menu-item-text {
--- a/theme/connector/view/_Learn.scss
+++ b/theme/connector/view/_Learn.scss
@@ -941,3 +941,32 @@ img.detail-has-data-very-small {
     text-transform: uppercase;
   }
 }
+
+/* Export button styles (for learn and data grids) */
+.x-btn-rounded-inverted-accent-small .x-btn-arrow-light {
+  background-image: url(data:image/gif;base64,R0lGODlhCgAKAHAAACH5BAEAAAIALAAAAAAKAAoAgQAAAP///wAAAAAAAAIMlI+pyxr/XFyh2ZsKADs=);
+  background-repeat: no-repeat;
+  background-position-x: 100%;
+}
+
+.export-item {
+  background-color: $white;
+
+  .x-menu-item-text {
+    margin-left: 10px;
+  }
+  .export-item-icon {
+    background-image: url(data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPCEtLSBHZW5lcmF0b3I6IEFkb2JlIElsbHVzdHJhdG9yIDIzLjAuNiwgU1ZHIEV4cG9ydCBQbHVnLUluIC4gU1ZHIFZlcnNpb246IDYuMDAgQnVpbGQgMCkgIC0tPgo8c3ZnIHZlcnNpb249IjEuMSIgaWQ9IkxheWVyXzEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiIHg9IjBweCIgeT0iMHB4IgoJIHZpZXdCb3g9IjAgMCA5IDEwIiBzdHlsZT0iZW5hYmxlLWJhY2tncm91bmQ6bmV3IDAgMCA5IDEwOyIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSI+CjxzdHlsZSB0eXBlPSJ0ZXh0L2NzcyI+Cgkuc3Qwe2ZpbGw6IzAwMDAwMDt9Cjwvc3R5bGU+CjxwYXRoIGNsYXNzPSJzdDAiIGQ9Ik05LDd2MnYxSDhIMUgwVjdoMXYyaDdWN0g5eiBNNy4zMywzLjk0TDYuNjIsMy4yM0w1LDQuODVWMEg0djQuODVMMi4zOCwzLjIzTDEuNjcsMy45NEw0LjUsNi43N0w3LjMzLDMuOTR6Ii8+Cjwvc3ZnPgo=);
+    background-repeat: no-repeat;
+    background-position-x: 0%;
+    background-size: 12px;
+  }
+}
+
+/* Inverted download background when the menu item is active */
+.export-item.x-menu-item-active .export-item-icon {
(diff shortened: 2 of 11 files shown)


theme/connector/lib/ext-02.css
```diff
@@ -6228,7 +6228,6 @@ x-boundlist-item {
     from(#fb4770),
     to(#fb2e5c)
   );
-  border-radius: 9px;
```
labkey-klum, line 6231: they wanted square,  not rounded button edges on hover

theme/connector/view/_Learn.scss
```diff
@@ -941,3 +941,33 @@ img.detail-has-data-very-small {
...
+  }
+}
+
+/* Inverted download background when the menu item is active */
+.export-item.x-menu-item-active .export-item-icon {
+  background-image: url(data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPCEtLSBHZW5lcmF0b3I6IEFkb2JlIElsbHVzdHJhdG9yIDIzLjAuNiwgU1ZHIEV4cG9ydCBQbHVnLUluIC4gU1ZHIFZlcnNpb246IDYuMDAgQnVpbGQgMCkgIC0tPgo8c3ZnIHZlcnNpb249IjEuMSIgaWQ9IkxheWVyXzEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiIHg9IjBweCIgeT0iMHB4IgoJIHZpZXdCb3g9IjAgMCA5IDEwIiBzdHlsZT0iZW5hYmxlLWJhY2tncm91bmQ6bmV3IDAgMCA5IDEwOyIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSI+CjxzdHlsZSB0eXBlPSJ0ZXh0L2NzcyI+Cgkuc3Qwe2ZpbGw6I0ZGRkZGRjt9Cjwvc3R5bGU+CjxwYXRoIGNsYXNzPSJzdDAiIGQ9Ik05LDd2MnYxSDhIMUgwVjdoMXYyaDdWN0g5eiBNNy4zMywzLjk0TDYuNjIsMy4yM0w1LDQuODVWMEg0djQuODVMMi4zOCwzLjIzTDEuNjcsMy45NEw0LjUsNi43N0w3LjMzLDMuOTR6Ii8+Cjwvc3ZnPgo=);
+  background-repeat: no-repeat;
+  background-position-x: 0%;
+  background-size: 12px;
+}
```
labkey-klum, line 975: admittedly, my CSS foo is pretty limited so I would be interested in hearing if there are ideas to handle this more efficiently

theme/connector/view/_Learn.scss
```diff
@@ -941,3 +941,33 @@ img.detail-has-data-very-small {
...
+
+/* Data export menu item, with download icon */
+.export-item {
+  background-color: $white;
+
+  .x-menu-item-text {
+    margin-left: 10px;
+  }
+  .export-item-icon {
+    background-image: url(data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPCEtLSBHZW5lcmF0b3I6IEFkb2JlIElsbHVzdHJhdG9yIDIzLjAuNiwgU1ZHIEV4cG9ydCBQbHVnLUluIC4gU1ZHIFZlcnNpb246IDYuMDAgQnVpbGQgMCkgIC0tPgo8c3ZnIHZlcnNpb249IjEuMSIgaWQ9IkxheWVyXzEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiIHg9IjBweCIgeT0iMHB4IgoJIHZpZXdCb3g9IjAgMCA5IDEwIiBzdHlsZT0iZW5hYmxlLWJhY2tncm91bmQ6bmV3IDAgMCA5IDEwOyIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSI+CjxzdHlsZSB0eXBlPSJ0ZXh0L2NzcyI+Cgkuc3Qwe2ZpbGw6IzAwMDAwMDt9Cjwvc3R5bGU+CjxwYXRoIGNsYXNzPSJzdDAiIGQ9Ik05LDd2MnYxSDhIMUgwVjdoMXYyaDdWN0g5eiBNNy4zMywzLjk0TDYuNjIsMy4yM0w1LDQuODVWMEg0djQuODVMMi4zOCwzLjIzTDEuNjcsMy45NEw0LjUsNi43N0w3LjMzLDMuOTR6Ii8+Cjwvc3ZnPgo=);
```
XingY: Can those svgs be saved in webapp/Connector/images/ as files and then used as file path?

theme/connector/view/_Learn.scss
```diff
@@ -941,3 +941,33 @@ img.detail-has-data-very-small {
     text-transform: uppercase;
   }
 }
+
+/* Data export button, with light arrow icon */
+.x-btn-rounded-inverted-accent-small .x-btn-arrow-light {
```
XingY: Looks like the spec is expecting the arrow to point up when menu is expanded. Is that handled as well?

theme/connector/view/_Learn.scss
labkey-klum: yes, but I thought it would be easier to use just the inlined data versions rather than check in two separate files that are only used in one place each. Do you know if there are pros/cons of one over the other?

theme/connector/view/_Learn.scss
labkey-klum: It's not right now, I was going to ask the client if that were a must have versus nice to have since it doesn't easily fall out of the Ext4 implementation. I'm sure I could do it but would add more CSS complexity.

theme/connector/view/_Learn.scss
labkey-klum: swtiched to the more traditional SVG files...

theme/connector/view/_Learn.scss
labkey-klum: implemented the up arrow when the menu is opened...

theme/connector/view/_Learn.scss
XingY: It's more of a personal preference. I found the file way to be more readable and consistent. For the "export" icon case, the inline method might have the advantage of being able to define a single css selector with the svg, but use additional selector to change just the color part of it. It looks good as is.

XingY (approved):

PR #491, opened 2022-07-09
FR 45774 : Update the study pages “Interactive report” section to allow for any URL

### LINKED ISSUE #489: Update the Study pages "Interactive report" section to allow for any URL
[Spec](https://docs.google.com/document/d/1l5no8D1UizAdquDCx9d9cs--c3K1SjrTkRfRyj2E6EE/edit#)

**Tasks**

- [x] Complete mini spec 0.5h
- [x] Impl plan -1.5 h @labkey-bpatel
- [x] Dev Review -0.5 h @labkey-klum 
- [x] Add more tasks -0.25 @labkey-bpatel
- [x] Add new columns to cds.studyReport and cds.import_studyReport - 1hr
- [x] Identify illegal combinations during an ETL Run - 2 hrs
- [x] Update sample data -  1 hr
- [x] Reconfigure Interactive Reports section - 2 hrs
- [x] Create a PR - 0.25 hr
- [x] Code Review - 1 hr
- [x] Code Review changes -  0 (adj hrs)
- [x] Update Spec - 0.5 hr
- [x] Create test plan - 1 hr
- [x] Manual Test - 2 hr @labkey-sweta 

**Hours**
IPH - 15
Kanban hr-13.5

labkey-bpatel:
#### Rationale
Support URLs for external reports in addition to the internal reports generated within CDS application on Learn about > Study pages.
[Spec](https://docs.google.com/document/d/1l5no8D1UizAdquDCx9d9cs--c3K1SjrTkRfRyj2E6EE/edit#)

#### Changes
* Schema updates to add new columns.
* Add a new transform task class to validate/identify illegal combination of values.
* Update 'Interactive Report' section for Learn about > Study pages to list external reports as well.

### DIFF
--- a/resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
+++ b/resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2022 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+TRUNCATE TABLE cds.import_studyReport;
+DROP TABLE cds.import_studyReport;
+
+CREATE TABLE cds.import_studyReport
+(
+    rowId SERIAL,
+    prot VARCHAR(250) NOT NULL,
+    cds_report_id INTEGER,
+    cds_report_link TEXT,
+    cds_report_label TEXT,
+    container ENTITYID NOT NULL,
+
+    CONSTRAINT PK_CDS_IMPORT_STUDY_REPORT PRIMARY KEY (rowId),
+    CONSTRAINT UQ_CDS_IMPORT_STUDY_REPORT UNIQUE (prot, cds_report_id, container, cds_report_link, cds_report_label)
+);
+
+TRUNCATE TABLE cds.studyReport;
+DROP TABLE cds.studyReport;
+
+CREATE TABLE cds.studyReport
+(
+    rowId SERIAL,
+    prot VARCHAR(250) NOT NULL,
+    cds_report_id INTEGER,
+    cds_report_link TEXT,
+    cds_report_label TEXT,
+    container ENTITYID NOT NULL,
+
+    CONSTRAINT PK_CDS_STUDY_REPORT PRIMARY KEY (rowId),
+    CONSTRAINT UQ_CDS_STUDY_REPORT UNIQUE (prot, cds_report_id, container, cds_report_link, cds_report_label)
+);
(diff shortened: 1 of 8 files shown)


resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
```diff
@@ -0,0 +1,47 @@
...
+    rowId SERIAL,
+    prot VARCHAR(250) NOT NULL,
+    cds_report_id INTEGER,
+    cds_report_link TEXT,
+    cds_report_label TEXT,
+    container ENTITYID NOT NULL,
+
+    CONSTRAINT PK_CDS_STUDY_REPORT PRIMARY KEY (rowId),
+    CONSTRAINT UQ_CDS_STUDY_REPORT UNIQUE (prot, cds_report_id, container, cds_report_link, cds_report_label)
+);
```
labkey-klum: Won't this force them to re-import all of their data after this upgrade? In the past wouldn't we just add new fields with reasonable defaults but preserve existing rows. It makes it a little trickier because you have to drop and then re-add constraints, see `cds-22.001-22.002.sql` for example.

labkey-klum (changes requested):
see comment on upgrade script and address if needed.

resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
labkey-bpatel: Correct it would force them to re-import, which they do anyway, dont' they normally?
I had implemented like you suggested previously, but ended up going this route because for some reason cds_report_id was showing up as a Required field even after dropping the PK constraint it was part of. And since they run ETLs to re-import data, thought it would be ok to drop & recreate

labkey-bpatel: From my local history, this is what i had done previously to drop the PK constraint, where cds_report_id was part of the PK ( 'PRIMARY KEY (prot, cds_report_id, container)'):

ALTER TABLE cds.import_studyReport DROP CONSTRAINT PK_ImportStudyReport; 
ALTER TABLE cds.studyReport DROP CONSTRAINT PK_studyReport;

And it dropped the constraints as expected, but cds_report_id kept showing up as a required field (i also tried truncating first)

labkey-klum: They only run the ETLs when they have new data that they want to expose in the app. Looking at their production server, it looks like the last time they ran that ETL was in March of this year.

That SQL looks correct, it would be good if we can get it to work without having to delete any rows but if that isn't possible we will need to instruct them to run that ETL on both deployments.

labkey-bpatel: Looks like column that's part of the PK gets set as NOT NULL by the db, once i dropped NOT NULL constraint on that column it now works as expected. 
Thanks!

resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+--  Update cds.import_studyReport table
+ALTER TABLE cds.import_studyReport ADD COLUMN cds_report_link TEXT;
+ALTER TABLE cds.import_studyReport ADD COLUMN cds_report_label TEXT;
```
labkey-klum: seems fine but do we anticipate needing to store information in excess of say 250 characters?

labkey-klum (approved):

resources/schemas/dbscripts/postgresql/cds-22.005-22.006.sql
labkey-bpatel: done

PR #516, opened 2022-09-22
Automation test Update export button and add learn pages export feature

labkey-sweta:
#### Rationale
Automation test Update export button and add learn pages export feature

#### Related Pull Requests
* 

#### Changes
New test class test/src/org/labkey/test/tests/cds/CDSLearnAboutExportTest.java

### DIFF
--- a/test/src/org/labkey/test/tests/cds/CDSGroupTest.java
+++ b/test/src/org/labkey/test/tests/cds/CDSGroupTest.java
@@ -495,6 +495,7 @@ public class CDSGroupTest extends CDSGroupBaseTest
 
     private void verifyExportButtonOnReports()
     {
+        cds.enterApplication();
         cds.viewLearnAboutPage("Reports");
         Locator.XPathLocator exportBtn = Locator.tagWithId("span", "learn-grid-export-button-id-btnIconEl");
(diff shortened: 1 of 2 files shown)


test/src/org/labkey/test/tests/cds/CDSGroupTest.java
```diff
@@ -488,6 +493,13 @@ private void verifyLinksOnStudyPage(String descr)
         verifyCuratedLink(descr);
     }
 
+    private void verifyExportButtonOnReports()
+    {
+        cds.viewLearnAboutPage("Reports");
```
labkey-bpatel, line 499: You might have to add a report, otherwise Reports tab will not be present if there are no reports; unless a test that creates a report always ran before this test.

labkey-sweta: VerifyExportButtonOnReports is added at the end of verifyInteractiveAndCuratedLinks() test, as per the test plan that test adds reports.
Am i missing something?

labkey-bpatel: Nope, I missed that.

Looks good!

labkey-bpatel (approved):

PR #533, opened 2022-12-06
[Dataspace] FR 46570 Add download links on learn assay pages

labkey-bpatel:
#### Rationale
This feature will allow users to export metadata associated with individual assays, mainly, BAMA, NAb, NAB MAb, and PK MAb assays and their respective Overview, Variables, and Antigens tabs (if applicable).

Spec [here](https://docs.google.com/document/d/1MEp05hW2P9e6iYvHLqVn90E-zp2Oy6gFg7TVTAQiX6U/edit#).

#### Changes
* Add export button to the Overview tab for BAMA, NAB, NAB MAb, PK MAb assays
* Add variables and antigen tab related metadata/qview for the columns to be exported
* Updates to CDSExportQueryView to handle Antigen export, and minor updates to handle metadata and variable export


webapp/Connector/src/view/Learn.js
```diff
@@ -132,6 +132,124 @@ Ext.define('Connector.view.Learn', {
         }
     },
 
+    requestAssayExportCSV : function(cmp, item, data) {
+        console.log("export csv");
```
labkey-klum: was this just debug logging that can be removed?

webapp/Connector/src/view/Learn.js
```diff
@@ -132,6 +132,124 @@ Ext.define('Connector.view.Learn', {
...
+        var qviewName = "AssayExportView";
+
+        //for antigen
+        var antigen_query = null;
+        if (data.assay_identifier.includes("BAMA")) {
+            antigen_query = "bamaantigen";
+        }
+        else if (data.assay_identifier.includes("NAB")) {
+            antigen_query = "nabantigen";
+        }
```
labkey-klum, line 163: Consider whether it would be worthwhile to generalize this setup based on new properties that we add to the assay metadata. It might make it easier in the future if they wanted to enable/disable exports in new or existing assays. The argument against this is that it would require them to modify and update their import metadata.

webapp/Connector/src/view/PageHeader.js
```diff
@@ -145,6 +147,42 @@ Ext.define('Connector.view.PageHeader', {
             }, this)
         });
 
+        if (this.model.data.assay_identifier) {
```
labkey-klum: Can this be switched to use configuration information for the components in `cube.js`? I was thinking about something similar to the `hasSearch` property for the details tabs in `cube.js:565`

labkey-klum (approved):

webapp/Connector/src/view/Learn.js
labkey-bpatel: Done

webapp/Connector/src/view/PageHeader.js
labkey-bpatel: Yup, good suggestion, now switched to use config info similar to hasSearch property

webapp/Connector/src/view/Learn.js
labkey-bpatel: Good point! I thought about making it more data driven than hard coded query names; since it would have been slightly more work than anticipated plus there were only two queries, i resorted to this shortcut way for now. 
I like your suggestion of adding it to the assay metadata - I've now included a comment along these lines. I also want to check with Drienna next week if there will be other antigens in the near future to make this change worthwhile.

PR #536, opened 2022-12-08
Adding additional ten mins to timeout to account for initial set up.

labkey-sweta:
#### Rationale
On a new database the first test sets up all CDS project which roughly takes 10 mins, since we cannot control the order in which tests run increasing the timeouts by 10 mins is the solution.

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):
Seems to have done the trick on TeamCity

PR #544, opened 2022-12-17
Ticket 46665: Homepage url redirects to wrong page when logged in

labkey-bpatel:
#### Rationale
Ticket [46665](https://www.labkey.org/Dataspace/Support%20Tickets/issues-details.view?issueId=46665): Homepage url redirects to wrong page when logged in

https://dataspace.cavd.org/ takes users to https://dataspace.cavd.org/project/home/begin.view instead of the application page.

#### Changes
* Add html page for redirecting users to the main application page. This html page should be set in Look and Feel Settings --> Alternative site welcome page: cds/pageRedirect.view

### DIFF
--- a/resources/views/pageRedirect.html
+++ b/resources/views/pageRedirect.html
@@ -1,4 +1,4 @@
 <head>
     <p>Redirecting to Dataspace application...</p>
-    <meta http-equiv="refresh" content="0; URL=https://dataspace.cavd.org/cds/CAVD/app.view" />
+    <meta http-equiv="refresh" content="0; URL=/cds/CAVD/app.view" />
 </head>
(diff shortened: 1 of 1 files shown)


resources/views/pageRedirect.html
```diff
@@ -0,0 +1,4 @@
+<head>
+    <p>Redirecting to Dataspace application...</p>
+    <meta http-equiv="refresh" content="0; URL=https://dataspace.cavd.org/cds/CAVD/app.view" />
```
labkey-jeckels: This will be confusing on staging servers and dev machines. Can we use a relative URL, like this?

```suggestion
    <meta http-equiv="refresh" content="0; URL=/cds/CAVD/app.view" />
```

labkey-jeckels (changes requested):

resources/views/pageRedirect.html
labkey-bpatel: It won't work on dev machines anyway, since the project and folder names are not the same.
In any case, Dataspace will be only using/care for this redirect for their production - it won't be set on their staging

labkey-jeckels (approved):

PR #571, opened 2023-04-11
Hide Learn Antigens tab if there is no data in the source table

labkey-bpatel:
#### Rationale
Client request to hide Learn about > Antigens tab if there is no data in the source table.

#### Changes
* Add a flag to display 'Learn about' tab only if there is data.
* Refactor hard-coded tab hiding for Reports and make it data driven.


labkey-klum (approved):
good refactor!

PR #568, opened 2023-03-26
[Dataspace] - New Table Spec for BCR Data

labkey-bpatel:
#### Rationale
Dataspace client will be getting data about the B-Cell Receptors (BCRs). These BCR data are key in further understanding of HIV infections and possibly bringing us one step closer to the cure/vaccine. Once the data is in LabKey, the client will be writing R scripts to manipulate the data using LabKey APIs.

Spec [here](https://docs.google.com/document/d/1zM6BVb5y4bGVJXDzMCV8N5jTDqe6QABwxE7zd1sq57I/edit#)

#### Changes
* Create new tables to store BCR data (note: I have not added all the FKs mentioned in the spec since the data provided seems incomplete causing FK constraint errors during import).
* Add table metadata
* Bump module version
* Add sample data
* Add support to import csv files

PR #573, opened 2023-04-25
[Dataspace] Add support to import and store B-Cell Receptor (BCR) data - Client Requested Changes

labkey-bpatel:
#### Rationale
Initial implementation for BCR tables was completed in [this](https://github.com/LabKey/cds/pull/568) PR. As the client refines the data and data usage, there will be updates to the existing table or addition of new tables and views - this PR covers such client requested changes.

Spec [here](https://docs.google.com/document/d/1_2m7oggaFsvrXOC809qYLBnLhaDWEy0t6TBdgFVkPZk/edit#heading=h.e7xmayx3y624).
#### Changes
* Updates to BCR tables
* Sample data updates
* ETL updates
* Add alignment and seq. germline queries


resources/queries/cds/alignment_mab_metadata.sql
```diff
@@ -0,0 +1,4 @@
+SELECT *
+FROM cds.alignment a
+    INNER JOIN cds.antibody_sequence aseq ON a.sequence_id = aseq.sequence_id
+    INNER JOIN cds.MAbMetadata mm ON aseq.mab_id = mm.mab_id
```
labkey-klum, line 4: I wasn't sure if you knew this but an `INNER` join is implied unless you specify the modifiers `left, right, outer, full`

resources/schemas/dbscripts/postgresql/cds-23.002-23.003.sql
```diff
@@ -0,0 +1,46 @@
+DROP TABLE cds.sequence_headers;
```
labkey-klum: could we have renamed the table, given that no field changes were made

labkey-klum (approved):

resources/queries/cds/alignment_mab_metadata.sql
labkey-bpatel: This is mainly for my readability, I normally use "inner join" instead of just "join"

resources/schemas/dbscripts/postgresql/cds-23.002-23.003.sql
labkey-bpatel: ha yes, i was just being lazy with copy pasting and removing "s" from 'headers' :)
but good to rename, i'll change it

PR #579, opened 2023-05-12
Ticket 47772: Improve performance when loading Learn about pages

labkey-bpatel:
#### Rationale
Ticket [47772](https://www.labkey.org/Dataspace/Support%20Tickets/issues-details.view?issueId=47772): Improve performance when loading Learn about pages

This PR addresses part of the lag where the queries are being loaded twice.

#### Changes
* Call updateFilters() only when the params are not empty

### DIFF
--- a/webapp/Connector/src/view/Learn.js
+++ b/webapp/Connector/src/view/Learn.js
@@ -836,7 +836,9 @@ Ext.define('Connector.view.Learn', {
             this.getHeader().on('selectdimension', this.loadDataView, this, {single: true});
         }
 
-        // this.getHeader().selectTab(dimension ? dimension.uniqueName : undefined, id, dimension, params);
+        if (params && Object.keys(params).length > 0) {
+            this.getHeader().selectTab(dimension ? dimension.uniqueName : undefined, id, dimension, params);
+        }
     }
 
 });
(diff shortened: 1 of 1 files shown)


webapp/Connector/src/view/Learn.js
```diff
@@ -830,12 +830,22 @@ Ext.define('Connector.view.Learn', {
...
         if (dimension) {
             this.loadDataView(dimension, id, urlTab, params);
+            this.getHeader().showExportButton(dimension);
         }
         else {
             this.getHeader().on('selectdimension', this.loadDataView, this, {single: true});
         }
 
-        this.getHeader().selectTab(dimension ? dimension.uniqueName : undefined, id, dimension, params);
+        if (params && Object.keys(params).length > 0) {
```
XingY: It seems the rewrite is to get rid of the updateFilters calls when params is not empty. Can we just add a param `skipFiltersUpdate` to the respective functions. I think that would be much cleaner.

`selectTab : function(dimUniqueName, id, dimension, params, skipFiltersUpdate)`
`filterStoreFromUrlParams: function(id, dimension, params, skipFiltersUpdate)`

labkey-bpatel: Maybe i missing something but the rewrite is to call updateFilters() when the params is not empty

XingY: Yep, sorry I meant to get rid of the updateFilters call when params is empty. so skipFiltersUpdate = !params || Object.keys(params).length === 0

labkey-bpatel: Got it! 
yes, much cleaner, now updated

XingY (approved):

PR #597, opened 2023-08-31
[DataSpace] FR 45527: Add link to static blogs on public page

labkey-bpatel:
#### Rationale
Add link to static blogs on public page. Other minor updates to the public page.
Spec [here](https://docs.google.com/document/d/1aYgnxAsJRPW8kTXnif8aQoU1gmaGsWdH2OeZ_qZIcQI/edit)

#### Changes
* Updates to get blog data on public page. 
* Add module properties for blog directory and all blogs.
* Update FrontPage to add new blog section and display four recent blogs.
* Convert images to thumbnail.
* Updates to the Welcome section to display the full 'down arrow' button.
* Updates to Always Growing section to use a pill background for stats.
* Styling updates.
* Add sample data.


src/org/labkey/cds/CDSController.java
```diff
@@ -191,12 +194,21 @@ public Object execute(Object o, BindException errors) throws Exception
             ApiSimpleResponse response = new ApiSimpleResponse();
             List<Item> items = new ArrayList<>();
 
-            List<RSSFeed> feeds = RSSService.get().getFeeds(getContainer(), getUser());
+            User user = getUser();
+
+            // In order to display blog posts on Dataspace's public page where the user is not required to log-in, we need to create a
+            // Limited user with Reader role otherwise no feeds will be returned by RSSService.get().getFeeds()
+            if (getUser().isGuest())
```
labkey-klum: I don't think we care if the user is a guest or not but rather if the user had read access to the folder:

```suggestion
            if (getContainer().hasPermission(user, ReadPermission.class))
```

src/org/labkey/cds/CDSController.java
```diff
@@ -191,12 +194,21 @@ public Object execute(Object o, BindException errors) throws Exception
...
+            {
+                user = new LimitedUser(User.guest, new int[0], Collections.singleton(RoleManager.getRole(ReaderRole.class)), false);
```
labkey-klum: We should also be able to reuse the current user object versus `User.guest`
```suggestion
                user = new LimitedUser(user, new int[0], Collections.singleton(RoleManager.getRole(ReaderRole.class)), false);
```

src/org/labkey/cds/CDSManager.java
```diff
@@ -330,6 +329,18 @@ public String getCDSImportFolderPath(Container container)
         return PageFlowUtil.decode(mp.getEffectiveValue(container));
     }
 
+    public String getBlogPath(Container container)
+    {
+        ModuleProperty mp = ModuleLoader.getInstance().getModule(CDSModule.class).getModuleProperties().get(CDSModule.BLOG_PATH);
+        return PageFlowUtil.decode(mp.getEffectiveValue(container));
```
labkey-klum: It would be nice if this method could tolerate either a relative or fully qualified URL but then return a fully qualified URL. That way the code in `frontPage.jsp` wouldn't need to concatenate prepend `baseURL` to the `blogPath`. For example the module property could be either:
- `/_webdav/home/Public/%40files/publicPage/blogs/`
- `http://localhost:8080/labkey/_webdav/home/Public/%40files/publicPage/blogs/`

but always include the base URL if necessary.

src/org/labkey/cds/view/template/FrontPage.jsp
```diff
@@ -73,6 +80,69 @@
...
+                            img.src = "<%=h(baseURL + blogPath)%>" + items[imageIdx].imageFile;
+
+                            console.log("img src = " + img.src);
+                            img.onload = function () {
+                                c.drawImage(img, 0, 0, canvas.width, canvas.height);
+                            }
+                            document.getElementById("thumbnail" + (imageIdx+1)).appendChild(canvas);
+                        };
+
+                        $('<td>').html('<a id="thumbnail1" href="' + items[0].link + '"></a>').appendTo($tr);
```
labkey-klum: Could we have a value for the max number of thumbnails to render and just loop over the news feed items to create the elements? It looks like this code will blow up if less than 4 items are returned from the news feed. 

Do we need to be more fault tolerant on whether all item properties may be present? Or is this validation happening elsewhere?

src/org/labkey/cds/view/template/FrontPage.jsp
```diff
@@ -73,6 +80,69 @@
...
+                            var canvas = document.createElement("canvas");
+                            canvas.width = 176.761;
+                            canvas.height = 100;
+                            canvas.style.border = "1px solid #DCDCDC";
+                            var c = canvas.getContext("2d");
+
+                            var img = new Image();
+                            img.src = "<%=h(baseURL + blogPath)%>" + items[imageIdx].imageFile;
+
+                            console.log("img src = " + img.src);
```
labkey-klum: left over debug code?

theme/front-page/index/sections.scss
```diff
@@ -20,7 +20,7 @@
   max-height: 62em;
 
   .fp-scrollable {
-    padding-top: 6em !important;
+    //padding-top: 6em !important;
```
labkey-klum: should we delete this?

theme/front-page/index/sections.scss
```diff
@@ -69,11 +69,11 @@
     @include tablet {
       margin: 3em 0 2.67em 0;
     }
-    @include legacy-desktop {
-      margin: 10.67em 0 2.67em 0;
-    }
+    //@include legacy-desktop {
```
labkey-klum: is this needed any longer?

labkey-klum (changes requested):
see comments with suggested changes.

src/org/labkey/cds/CDSController.java
labkey-bpatel: getContainer().hasPermission(user, ReadPermission.class) evaluates to false. Guest user doesn't have a read permission, hence the reason to create a LimitedUser with ReaderRole.

src/org/labkey/cds/CDSController.java
labkey-bpatel: Good point. Done!

theme/front-page/index/sections.scss
labkey-bpatel: Removed

theme/front-page/index/sections.scss
labkey-bpatel: Deleted

src/org/labkey/cds/view/template/FrontPage.jsp
labkey-bpatel: Cleared

src/org/labkey/cds/CDSManager.java
labkey-bpatel: Done

src/org/labkey/cds/view/template/FrontPage.jsp
labkey-bpatel: Now iterating through max number of items.

Did not make it fault tolerant (or initially add any checks) since these blog items are already in use to display on the Dataspace's home page, and there already are more than 4 blogs present, and new ones will always get added. And if the image file is not found or the thumbnail property is absent in the json file then it will show an empty box like so:
image

src/org/labkey/cds/CDSController.java
labkey-klum: sorry missed a `!` in front of that conditional it should have read more like:
```
if (!getContainer().hasPermission(user, ReadPermission.class))
```
Unless we are really trying to distinguish between guest and non-guest users, checking the access level is the preferred route.

labkey-bpatel: Ah, got it! That makes sense, thank you for clarifying. Now updated!

labkey-klum (approved):

PR #602, opened 2023-10-23
Various issues with the CDS application

labkey-klum:
#### Rationale
- [48892: MAb data grid loading mask fails to clear
](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48892) This was fixed by following a similar pattern we use in the learn pages to avoid the case where the data may get loaded before the loading mask delay.
- [48894: CDS : layout issues on the filter info pane](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48894)
- [48916: CDS : search not working for assay variables or antigens](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48916) Interesting bug here, the fix is in `LearnGrid.js` and it has to do with how class level defined listeners interact with classes that may extend the class. Both `AssayAntigen.js` and `VariableList.js` extend this class and define `isDetailedLearnGrid = true` in order to initialize the details grids in the sub tabs of the assay details view. The override doesn't seem to work and the `this` variable in the closure is different from the one in either of the two derived classes. I don't understand completely why this is the case but to fix it I had to move the initialization of the listeners into `initComponent`.
- [48938: CDS - frontpage notification area is encoding HTML](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48938)

### DIFF
--- a/theme/connector/view/_InfoPane.scss
+++ b/theme/connector/view/_InfoPane.scss
@@ -2,8 +2,6 @@
   background-color: $white;
   font-size: 12px;
   padding: 24px 16px!important;
-  margin: 6px!important;
-  width:230px!important;
   @include prefixer(border-radius, 2px);
   @include prefixer(box-shadow, 0 1px 0 0 $gray-6);
 
--- a/webapp/Connector/src/plugin/Messaging.js
+++ b/webapp/Connector/src/plugin/Messaging.js
@@ -118,6 +118,7 @@ Ext.define('Connector.plugin.Messaging', {
                 x: this.calculateX(this, box, msg),
                 y: this.calculateY(this, box, msg),
                 listeners: listeners,
+                width: 150,
                 keep: keep,
                 modal: modal === true,
                 scope: this
--- a/webapp/Connector/src/view/Learn.js
+++ b/webapp/Connector/src/view/Learn.js
@@ -541,7 +541,7 @@ Ext.define('Connector.view.Learn', {
                 }
             }
             else {
-                if(store.getCount() === 0)
+                if(store.getCount() === 0 && !store.isFiltered())
                     store.loadSlice();
                 else
                     this.sortAndFilterStoreDelayed(store);
--- a/webapp/Connector/src/view/PageHeader.js
+++ b/webapp/Connector/src/view/PageHeader.js
@@ -134,7 +134,7 @@ Ext.define('Connector.view.PageHeader', {
             flex: 2
         };
 
-        var detailSearchField = Ext.create('Ext.form.field.Text', {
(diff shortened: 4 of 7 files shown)


theme/connector/view/_InfoPane.scss
```diff
@@ -2,8 +2,6 @@
   background-color: $white;
   font-size: 12px;
   padding: 24px 16px!important;
-  margin: 6px!important;
-  width:230px!important;
```
labkey-klum, line 6: This was for the info pane layout issues. I'm not sure why these overrides were added in the first place instead of just letting the component lay itself out within it's containers.

webapp/Connector/src/plugin/Messaging.js
```diff
@@ -118,6 +118,7 @@ Ext.define('Connector.plugin.Messaging', {
                 x: this.calculateX(this, box, msg),
                 y: this.calculateY(this, box, msg),
                 listeners: listeners,
+                width: 150,
```
labkey-klum, line 121: This fixes the button overflow issue when clearing the active filter. In other places we typically provide a width when using an Ext.Window component and this seems to make the element size more accurately on chrome and safari.

webapp/Connector/src/view/Learn.js
```diff
@@ -541,7 +541,7 @@ Ext.define('Connector.view.Learn', {
                 }
             }
             else {
-                if(store.getCount() === 0)
+                if(store.getCount() === 0 && !store.isFiltered())
```
labkey-klum, line 544: Needed to correctly handle events where the store was being filtered, versus being uninitialized.

webapp/frontPage/js/application.js
```diff
@@ -51,7 +51,7 @@ require(['jquery', 'scroll', 'modal', 'util'], function( $, scroll, modal, util)
             const msgDiv = $('div.notification-messages');
 
             $.each(o.messages, function (idx, msg) {
-              msgDiv.append($('<span>').text(msg).append($('<br>')));
+              msgDiv.append($('<span>').append(msg).append($('<br>')));
```
labkey-klum, line 54: The `text` method always encodes content.

webapp/Connector/src/view/PageHeader.js
```diff
@@ -134,7 +134,7 @@ Ext.define('Connector.view.PageHeader', {
             flex: 2
         };
 
-        var detailSearchField = Ext.create('Ext.form.field.Text', {
+        var detailSearchField = Ext.create('Connector.view.Learn.SearchField', {
```
labkey-klum, line 137: I did this to standardize the search field for all the learn pages, the custom control will render the clear (x) icon when there is search filter present.

labkey-bpatel (approved):

PR #603, opened 2023-11-06
Enable the learn tabs to wrap to multiple lines

labkey-klum:
#### Rationale
We started noticing test failures related to the new `groups` tab that was added to the learn pages. While we couldn't consistently reproduce it locally the tests showed the `publications` tab wrapping to the second line but overlapping with the column headers on the learn grid.

This change forces the learn header to wrap the learn tabs when there isn't enough horizontal space to render all of the containing elements. Most of this was accomplished using `CSS` styling, but some adjustments had to be made to remove explicit heights for learn header elements so the layout manager could position container elements appropriately. 

If you are trying this PR out locally keep in mind you'll need to run `npm run build` to generate the theme stylesheets.

image

### DIFF
--- a/theme/connector/view/_Learn.scss
+++ b/theme/connector/view/_Learn.scss
@@ -101,6 +101,12 @@ div.learnview {
   }
 }
 
+.learnheader {
+  display: flex;
+  display: -webkit-flex;
+  flex-wrap: wrap;
+}
+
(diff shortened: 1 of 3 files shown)


labkey-bpatel (approved):
Looks great! Thank you for fixing this.

PR #608, opened 2023-11-13
Render n levels of treatment group nesting

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49048)

The groups section in the study details learn pages are sourced by content from the `groups` field from the `cds.metadata.study` table. This content is in `HTML` format and is typically an unordered list of treatment groups and study ARMs. In order to limit the number of top level groups that are rendered on the page we parse the raw HTML to get the array of `li` items. The problem was that the previous code assumed only a single level of list items, but some of the protocols have nested sub-lists up to 2 levels deep: 

image

This PR introduces a change to support an arbitrary number of nesting levels (provided the HTML is reasonable well formed). And then renders those nesting levels in the details page while preserving the feature which limits the number of list items on the page (10). The limitation is applied only to the number of top level list items and does not count any of the sub nesting levels.

### DIFF
--- a/webapp/Connector/src/app/store/StudyOverview.js
+++ b/webapp/Connector/src/app/store/StudyOverview.js
@@ -587,26 +587,32 @@ Ext.define('Connector.app.store.StudyOverview', {
     },
 
     parseGroupList : function(groupHtml){
+        /**
+         * Recursively search for child list items and attach them to
+         * the parent object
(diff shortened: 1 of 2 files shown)


labkey-bpatel (approved):
Looks great!
Also, verified.

PR #621, opened 2023-12-28
[Dataspace] - FR 48359: Curated groups and plots improvements - Active Filters

labkey-bpatel:
#### Rationale
 The 'Active Filter' panel keeps track of all the filters applied by a user during their session. The aim of this story is to enhance the workflow for creating "Active Filters" (aka subject Groups) to make it easier and more intuitive for users to create custom groups. The workflow includes creating a new group by applying the filters or editing an existing group/filter.

Spec [here](https://docs.google.com/document/d/1v9Hk58ys0P9UVdWXYTvuYWL-0NMS994-Tgpy295l3YM/edit).

Items not implemented in this fb are listed [here](https://docs.google.com/document/d/1v9Hk58ys0P9UVdWXYTvuYWL-0NMS994-Tgpy295l3YM/edit#heading=h.rb0bktc5lgkf) in the spec.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5087

#### Changes
* Add 'Save as a group' button when filters are first created.
* Handle 'Save as a group' button click by displaying Group Save panel in the same region as Active filters section.
* Add 'Cancel' and 'Save group' buttons and handlers to save a new group.
* Add 'Cancel' and 'Save' menu buttons with 'Update this group' and 'Save as new group' menu items.
* Display group label on top of the saved filter/group.
* Update existing Save and Update group code for saving and updating.
* Hide and show buttons and components depending where the user is in the workflow.
* Take user to the group details page upon saving and updating the group.
* Apply styling updates.


webapp/Connector/src/controller/FilterStatus.js
```diff
@@ -30,6 +30,7 @@ Ext.define('Connector.controller.FilterStatus', {
                 navfilter.add({
                     xtype: 'container',
                     itemId: 'filterstatuscontainer',
+                    id: 'filterstatuscontainer-id',
```
labkey-klum, line 33: Just curious, what are we using the ID's on these elements for?

webapp/Connector/src/view/FilterStatus.js
```diff
@@ -15,11 +15,80 @@ Ext.define('Connector.view.FilterStatus', {
...
+                this.hideGroupSavePanel();
+                this.resetFilterStatusBackGroundColor();
+                this.hideEditGroupBtn();
+                this.hideSaveAsGroupBtn();
+            }
+        }
+    },
+
+    hideGroupLabel() {
+        Ext.getCmp('savedgroupname-id').hide();
```
labkey-klum, line 38: Alternatively, if you are searching within the hierarchy of this container you can use the `itemId` with `Container.getComponent`.

webapp/Connector/src/view/FilterStatus.js
```diff
@@ -68,22 +137,131 @@ Ext.define('Connector.view.FilterStatus', {
...
+            cls: 'savedgroup-label-container',
+            hidden: true,
+            layout: {
+                type: 'hbox'
+            },
+            items: [{
+                xtype: 'box',
+                cls: 'savedgroup-label',
+                id: 'savedgroup-label-id',
+                tpl: new Ext.XTemplate(
```
labkey-klum, line 183: This doesn't need to be a template instance, you can also pass back the string literal definition since everything here is a config (not an object instance).

labkey-klum (approved):

webapp/Connector/src/controller/FilterStatus.js
labkey-bpatel: To debug the dom, otherwise ext assigns its own id which makes it hard to debug at times

webapp/Connector/src/view/FilterStatus.js
labkey-bpatel: Ah I see, good to know! thanks!

PR #624, opened 2024-01-03
Automation test for Password strengthening rules

labkey-sweta:
#### Rationale
Automation test for Password strengthening rules

#### Related Pull Requests
* 

#### Changes
*


test/src/org/labkey/test/components/cds/ChangePasswordDialog.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String getErrorMessage()
+    {
+        return newElementCache().notification.getText();
+    }
+
+    public void submit()
+    {
+        newElementCache().submitButton.click();
+    }
```
labkey-tchad: `submit` should wait for the dialog to disappear or go stale. There should be a separate `submitExpectingError` method for the error test cases to use.

labkey-tchad (changes requested):
Is it really necessary to make everything in `ChangePasswordDialog` refinding? We should be able to know when staleness happens and wait for it explicitly.

labkey-klum (approved):

labkey-tchad (approved):
Looks good but failing on TeamCity

PR #629, opened 2024-01-14
Ticket 49336: New Groups in Learn - Groups not showing in Learn

labkey-bpatel:
#### Rationale
Client filed issues as mentioned in Ticket [49336](https://www.labkey.org/Dataspace/Support%20Tickets/issues-details.view?issueId=49336):
- Groups displaying on Home page but not in Learn
- Groups not displaying on Study page
- Clicking on a Group from the Home page takes user to a different Group details page than expected.

#### Changes
* Fix for groups not displaying on Learn pages.
* Query update (studyCuratedGrpWithLabel.sql) to display Curated groups on Study pages, instead of having to update studyCuratedGroup table each time a new group is created, which is going to have different group id on staging and production and client would have to update tables on both staging and prod and run an ETL for the data to get populated - this query update eliminates all these steps.
* Get the label from the clicked item and find the corresponding record based on the label instead of using Ext's auto mechanism of using store index.


labkey-bpatel:
Closing this PR, new 24.1 fb and PR (https://github.com/LabKey/cds/pull/631) created targeting 24.1 release

PR #643, opened 2024-02-15
Test updates

labkey-sweta:
#### Rationale
Fixing some random test failures on teamcity.

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95/2906125?buildTab=overview&expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/test/src/org/labkey/test/components/cds/ActiveFilterDialog.java
+++ b/test/src/org/labkey/test/components/cds/ActiveFilterDialog.java
@@ -2,6 +2,7 @@ package org.labkey.test.components.cds;
 
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
+import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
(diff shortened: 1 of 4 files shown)


test/src/org/labkey/test/components/cds/ActiveFilterDialog.java
```diff
@@ -82,18 +83,31 @@ public void saveGroup()
...
         {
-            getWrapper().waitForElement(elementCache().errorMsg);
+            count++;
+            try
+            {
+                elementCache().saveGroup.click();
+                errMsg = elementCache().errorMsg.findElement(_activeFilterDialogEl).getText();
+                worked = true;
+            }
+            catch (NoSuchElementException ex)
```
labkey-danield, line 100: Is this to catch the 'saveGroup' button not being there or the 'errorMsg' not being there? When I look at the code it has replaced I'm not completely sure which one was the issue.

Ideally you wouldn't want to have a retry like this (although they do exist in other test code). If this is working I'm wondering if part of the success comes from the sleep in the catch. Have you tried explicitly moving the mouse to the button, sleeping for the 1/2 second then doing the click? I know the click should automatically move the mouse to the button but it might be worth trying.

labkey-danield (commented):
@labkey-sweta maybe we can sync up and you can help me understand the retry code. We might be able to find a more reliable alternative.

labkey-klum (approved):
password changes seem fine, maybe sync up with Dan for his suggested changes?

PR #645, opened 2024-02-22
CDS learn group page refactor

labkey-klum:
[spec](https://docs.google.com/document/d/1kWJ_k5NxyOn1f2Fh4pNUs5gq2eCS0hPnkd5g4pmFCwo/edit#heading=h.bcr61ba681l)

#### Rationale
This PR pivots the source of the learn grid away from being sourced from queries from the study and cds tables and now sources them from the individual filters that are serialized as part of the saved group. This was the primary problem that was discovered where the groups that appeared in the home page listing differed from the groups that appeared in the learn grid. Additionally, information about the group in the learn grid could also be inaccurate where studies (or other column information) were associated with the group that was not saved in the initial filter.

#### Changes
- Source the study, assay, species and product information directly from the group filters. These are serialized in a JSON format and new code was added to parse the levels and members.
- Consolidate the group model and store:
  - There were two models and stores in the application, one for the group home page listing and the other for the group learn grid.
  - This was causing some problems when synchronizing the views after the group store was mutated.
  - Fields were unified and consolidated
  - The `Connector.model.Group` store was deleted. This was formerly the model and store for the home page listing. Now everything uses the newer `Connector.app.model.Group` model and associated store.


labkey-bpatel (commented):
Looks great overall. Now the newly saved group is reflected in the Learn grid, yay!

I did some manual testing, noticed a couple of things:
1. For me, Learn groups is displaying 'No related species' and 'No related products' for species and products even though there are species and products associated with the group (counts that are displayed below the Active filters shows Species 1,  Products 3, for ex.)
3. Find subjects > by Assays > select an Assay > Save as group.. > In Learn Groups, I don't see Assays listed for that group, not sure where we landed on that, to show Assays or not to show Assays. Also, going to Group Summary, it displays the Assay under Studies, instead of studies associated with that assay.

labkey-klum:
> 1. For me, Learn groups is displaying 'No related species' and 'No related products' for species and products even though there are species and products associated with the group (counts that are displayed below the Active filters shows Species 1,  Products 3, for ex.)

I wouldn't expect species to get pulled into the grid unless it's explicitly on the filter. 

> 2. Find subjects > by Assays > select an Assay > Save as group.. > In Learn Groups, I don't see Assays listed for that group, not sure where we landed on that, to show Assays or not to show Assays. Also, going to Group Summary, it displays the Assay under Studies, instead of studies associated with that assay.

We should show the assay if you find subjects by assay type. Good catch on the Assay by study which can potentially have both study and assay information, I'll get that fixed.

labkey-klum:
@labkey-bpatel I pushed some changes to help with your second comment.

labkey-bpatel:
> > 1. For me, Learn groups is displaying 'No related species' and 'No related products' for species and products even though there are species and products associated with the group (counts that are displayed below the Active filters shows Species 1,  Products 3, for ex.)
> 
> > > I wouldn't expect species to get pulled into the grid unless it's explicitly on the filter.
> > > > Got it, I thought so as well, but wanted to verify.
> 
> > 2. Find subjects > by Assays > select an Assay > Save as group.. > In Learn Groups, I don't see Assays listed for that group, not sure where we landed on that, to show Assays or not to show Assays. Also, going to Group Summary, it displays the Assay under Studies, instead of studies associated with that assay.
> 
> > > We should show the assay if you find subjects by assay type. Good catch on the Assay by study which can potentially have both study and assay information, I'll get that fixed.
> > > > Verified, looks good.

labkey-bpatel (approved):

PR #649, opened 2024-04-18
Create new CdsGrid component with better scoping and synchronization

labkey-tchad:
#### Rationale
Poor scoping of grids helpers causes tests to find hidden grids sometimes. Poor synchronization of grid operations forces tests to sleep to wait for operations to complete, which leads to race conditions in tests.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1907

#### Changes
* New `CdsGrid` component
* Make `LearnGrid`, `DataGrid`, and `MabDataGrid` wrap `CdsGrid` for better synchronization

### DIFF
--- a/test/src/org/labkey/test/components/cds/BaseCdsComponent.java
+++ b/test/src/org/labkey/test/components/cds/BaseCdsComponent.java
@@ -1,0 +1,29 @@
+package org.labkey.test.components.cds;
+
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+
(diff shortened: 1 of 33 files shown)


test/src/org/labkey/test/components/cds/BaseCdsComponent.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public final WebDriver getDriver()
+    {
+        return _wdw.getDriver();
+    }
+
+    @Override
+    public final WebDriverWrapper getWrapper()
+    {
+        // Avoid creating a new `Ext4Helper`, which resets the css prefix
```
labkey-sweta, line 26: How is this comment relevant here?

labkey-sweta (approved):

PR #656, opened 2024-07-30
Fix for StaleElementReferenceException

labkey-sweta:
#### Rationale
Trying to fix the intermittent test failure on teamcity. 

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95/3127006?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false&expandBuildTestsSection=true

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/test/src/org/labkey/test/pages/cds/LearnDetailsPage.java
+++ b/test/src/org/labkey/test/pages/cds/LearnDetailsPage.java
@@ -19,6 +19,7 @@ import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.cds.BaseCdsComponent;
 import org.labkey.test.util.LogMethod;
+import org.labkey.test.util.cds.CDSHelper;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
I suspect that the problem is with `LearnDetailsPage.clickBack`. It probably needs to wait for the grid to refresh, similar to [`CDSHelper.NavigationLink.LEARN.makeNavigationSelection`](https://github.com/LabKey/cds/blob/develop/test/src/org/labkey/test/util/cds/CDSHelper.java#L1641)

If you don't want to wire up `LearnDetailsPage.clickBack` to handle the grid refresh, I suggest pausing (`WebDriverWrapper.sleep(CDS_WAIT);`) at the end of `clickBack` if the search input has a value.

labkey-tchad (approved):

labkey-danield (approved):

PR #657, opened 2024-08-22
Simplify import from file ETL task

labkey-klum:
#### Rationale
The change refactors the ETL task used to import the initial data into a Dataspace instance. The ETL takes tabular data from specific files and imports them into tables in the `cds` schema.

This update addresses some specific problems:
- Certain tables expected `csv` formatted data while others expected `tsv` data. This made it difficult for the client to know which format to save per table. We no longer care as long as the file is one of our standard formats : `xls, xlsx, csv, tsv, txt`.
- No longer need the specific delimiter variants of `CDSImportCopyConfig`, so `CSVCopyConfig` and `TSVCopyConfig` can be deleted.
- Cleaned up the confusing array of constructor variants. There are only two now and the most that is needed is the import file name and the target query name.


src/org/labkey/cds/data/CDSImportCopyConfig.java
```diff
@@ -81,17 +100,17 @@ public DataIteratorBuilder selectFromSource(Container container, User user, Data
         }
         else
         {
-            File file = getByExtension(dir, getFileExtension(), ".txt");
-            if (null == file || !file.exists())
+            _importFile = getImportFile(dir);
+            if (null == _importFile || !_importFile.exists())
             {
-                context.getErrors().addRowError(new ValidationException("Could not find data file: \'" + _fileName + "\' (" + getFileExtension() + ", .txt)."));
+                context.getErrors().addRowError(new ValidationException("Could not find data file: '" + _fileName + "' (xls, xlsx, csv, tsv, txt)."));
```
labkey-bpatel: Could you use VALID_EXTENSIONS instead of (xls, xlsx, csv, tsv, txt)?

src/org/labkey/cds/data/CDSImportCopyConfig.java
```diff
@@ -33,37 +33,56 @@
...
 
+    public File getImportFile()
+    {
+        return _importFile;
+    }
+
+    private DataIteratorBuilder getDataLoader(File file) throws IOException
+    {
+        DataLoaderFactory factory = DataLoaderService.get().findFactory(file, null);
+        return factory.createLoader(file, true);
```
labkey-bpatel, line 83: Nice, this way is so much cleaner!

src/org/labkey/cds/data/CDSImportCopyConfig.java
```diff
@@ -33,37 +33,56 @@
...
-     * @param copyToImportTable Copy into an intermediate table with an import_ prefix, this can be useful if further transformation
-     *                          needs to be performed before copying to the final destination. Otherwise, set false to
-     *                          copy directly into the destination table.
+     * Constructor which imports a text file into a table in the CDS schema.
+     * @param targetQuery the target table
+     * @param fileName the name of the file, without the extension
+     */
+    public CDSImportCopyConfig(String targetQuery, String fileName)
+    {
+        super(null, null, "cds", targetQuery);
```
labkey-bpatel, line 61: Since we're setting sourceSchema and sourceQuery to null via super(null, null...) here, on lines 91 and 97, any calls to this.getSourceSchema() and getSourceQuery() will return null. It seems that the code block from lines 89-100 may no longer be relevant due to your changes, but further testing is needed to confirm this.

labkey-bpatel (approved):
This looks great! Please review my comments. Thank you!

src/org/labkey/cds/data/CDSImportCopyConfig.java
labkey-klum: agreed, done.

PR #681, opened 2025-01-27
Updating the reports test

labkey-sweta:
#### Rationale
Waiting longer for "View data" to load.
image

Link to teamcity test failure https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_CdsPostgres95/3352480

The fix is for CDSRReportsTest.validateViewReportBasic and CDSRReportsTest.validateViewReportComplex

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):
Might be worth consulting Karl as to whether it is acceptable for the schema browser to take >10 seconds to load. It takes around 5 seconds for me locally but only if I clear caches before hitting the schema browser.

labkey-klum (approved):

